PR #2, opened 2017-01-28
Feature/python3 support

### LINKED ISSUE #1: Python 3 support
Hello, is there a road-map to implement Python 3 support?

gatesn:
Adding experimental Python3 support. Confirmed working against VSCode.

Fixes #1

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,7 @@ addopts =
     --cov pyls --cov test
 
 [tox]
-envlist = py27,py35,py36
+envlist = py27,py36
 
 [testenv]
 commands = py.test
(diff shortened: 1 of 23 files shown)


tox.ini
```diff
@@ -14,7 +14,7 @@ addopts =
     --cov pyls --cov test
 
 [tox]
-envlist = py27,py35,py36
+envlist = py27,py36
```
MathieuDuponchelle: Would it not be better to test against 3.4 ? The test suite passes here, and 3.6 is not yet widely available.

gatesn: Yup fair point.

Issue #15, opened 2017-03-05
[Suggestion] Use File based logger


HerringtonDarkholme:
Hi, thanks for bringing us Python Language Server!

I'm using it with LanguageClient_neovim. And I'm debugging some issue with PLS.
Usually, the server is opened by language client. Users cannot control where language server can log.

PLS now logs to stderr by default. It would be helpful if PLS can log to a file. So editor users can inspect log independent of language client's behavior. For example, logging to `/tmp/LanguageServer.log` by default and file path can be configured by command line argument.

Thank you!

gatesn:
Will look to implement something similar to here: https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/

Sensible defaults but read logging config from file if it's available

HerringtonDarkholme:
Actually this problem is more serious. https://github.com/autozimu/LanguageClient-neovim/issues/21

It seems writing to stderr blocks writing to stdout.

gatesn:
So the problem, as you pointed out, is that languageClient-neovim sets `stderr=PIPE`.  What is happening is the client blocks waiting to read from stdout, Pyls fills up stderr buffer and blocks waiting to write to stderr and we end up in deadlock.

See here for more explanation:
http://stackoverflow.com/questions/1180606/using-subprocess-popen-for-process-with-large-output

What languageClient-neovim _should_ be doing is use select to read off both stdout and stderr and then throwing away stderr if it doesn't think its relevant. Alternatively it could set stderr=None to just drop it entirely, it could redirect it to appear on the client's stderr, or it could pass in a file object to redirect stderr to disk.

I agree that Pyls should have more configurable logging and will work on implementing that, but it's not in any language server contract that logs cannot be written to stderr.

Equally changing the log level isn't sufficient as in the case there is a lot of `ERROR` messages it will still hang! Just happening less frequently. 

So I think the solution here is for us to support configurable logging, and to make things easy for clients maybe a command line switch to redirect stderr to a file, but also for languageClient-neovim to properly handle stderr.

HerringtonDarkholme:
I don't think LanguageClient is the same as the stackoverflow question.

Actually PYLS blocks at writing to stdout, not stderr. Recorded here autozimu/LanguageClient-neovim#21

Source:
https://github.com/palantir/python-language-server/blob/master/pyls/jsonrpc.py#L155

gatesn:
[...]

PR #16, opened 2017-03-06
Support change loglevel from command line.

autozimu:
Also, set default loglevel to "ERROR". As right now, logger output
messages to stderr.

### DIFF
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -21,6 +21,10 @@ def add_arguments(parser):
         "--port", type=int, default=2087,
         help="Bind to this port"
     )
+    parser.add_argument(
+        "--loglevel", default="ERROR",
+        help="Set logging level"
+    )
 
 
 def main():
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @autozimu! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

autozimu:
A little background https://github.com/autozimu/LanguageClient-neovim/issues/21.

Basically the issue is because pyls write huge amount of logging messages to stderr, exhausting the stderr buffer. When using Popen to open pyls as subprocess, the piped stderr is buffered, and that buffer size usually has a limit of 4K.

gatesn:
Thanks for this @autozimu and PRs are always more than welcome! However I don't believe in this case changing the level is sufficient, it just postpones the issue instead of fixing it, c.f. #15

gatesn:
Closed in favour of #19

PR #19, opened 2017-03-07
Allow logging to be configurable from file or simply redirect logs to disk

### LINKED ISSUE #15: [Suggestion] Use File based logger

gatesn:
Also the default log level is changed to WARNING. Resolves #15

### DIFF
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -1,14 +1,18 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import argparse
+import json
 import logging
-import time
+import logging.config
 import sys
 from . import language_server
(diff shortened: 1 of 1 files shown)


HerringtonDarkholme:
Thanks! Looks good to me!

Issue #17, opened 2017-03-07
[Feature Request] More performant server


HerringtonDarkholme:
Hi. Python language server is well written, thank you for your contribution.

However, after using I find some improvements can be done.

1. logger 
Tracked by https://github.com/palantir/python-language-server/issues/15. Logging to stderr will block stdout writing.

2. threaded reading
Now PLS in stdio mode is single threaded. https://github.com/palantir/python-language-server/blob/6c781d4012521017e6a84e6d7d49827644ae669e/pyls/language_server.py#L29
So if client sends too many requests than stdin can buffer, PLS will also block because it cannot write to stdout. (It seems stdin/stdout/stderr share one common buffer)

3. cached compeletion
PLS generally is slower than deoplete-jedi, another jedi based completion plugin. The main reason is deoplete-jedi uses a cache. https://github.com/zchee/deoplete-jedi/blob/master/rplugin/python3/deoplete/sources/deoplete_jedi/cache.py If PLS can also support cache it should be faster.

Thanks!

gatesn:
+1 to all three here!

HerringtonDarkholme:
Hi @gatesn! Thanks for quick reply.

Currently I use PLS with neovim. But I would like to try it out with VSCode like that in README. However I cannot find extension in VSCode market. Would you like provide a installation manual?

Or do I need create a wrapper extension for PLS? Thanks!

gatesn:
@HerringtonDarkholme nice idea. I've noticed a few language servers do come with wrapper extensions. Would be happy to look into it some time, could you create a separate issue for it? Thanks

gatesn:
@HerringtonDarkholme thanks for raising these points. I want to close this ticket and instead track individually any ideas for performance improvements. Here are some to get us started:

https://github.com/palantir/python-language-server/labels/performance

PR #21, opened 2017-03-07
add threaded writing

HerringtonDarkholme:
As suggested in #17 

Threaded writing will prevent bad client side blocking.

### DIFF
--- a/pyls/jsonrpc.py
+++ b/pyls/jsonrpc.py
@@ -3,8 +3,11 @@ import json
 import logging
 import re
 import socketserver
+import threading
+from collections import deque
 
 log = logging.getLogger(__name__)
+cv = threading.Condition()
 
 
 class JSONRPCError(Exception):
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @HerringtonDarkholme! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

guyzmo:
Hi! Just a thought on this: wouldn't it be better to avoid using threading for that task but rather rely on simpler monothreaded concurrent programming? (using some sort of sugar around select/epoll/kqueue like what asyncio does).

All in all, I just started using the tool and it's ♥

gatesn:
I agree that threaded writing is not the right solution here, given the rest of the server is currently single-threaded it shouldn't be necessary. 

I believe the right approach here might be based around a concurrent results queue that each request writes back to with another thread consuming and sending out the results.

Issue #39, opened 2017-04-06
TCP mode does not work


MathieuDuponchelle:
Simplest client possible:

```
import requests
import json


def main():
    url = "http://localhost:4000/"
    headers = {'content-type': 'application/json'}

    payload = {
        "method": "initialize",
        "jsonrpc": "2.0",
        "id": 0,
    }
    response = requests.post(
        url, data=json.dumps(payload), headers=headers).json()

if __name__ == "__main__":
    main()
```

Run pyls with:

```
pyls --tcp --host 0.0.0.0 --port 4000
```

Result:

```
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 52078)
Traceback (most recent call last):
  File "/usr/lib64/python3.4/socketserver.py", line 617, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib64/python3.4/socketserver.py", line 344, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/meh/devel/sandbox/python-language-server/pyls/python_ls.py", line 33, in __init__
    super(PythonLanguageServer, self).__init__(*args, **kwargs)
TypeError: __init__() takes 3 positional arguments but 4 were given
----------------------------------------
```

MathieuDuponchelle:
After looking at the source code, it is pretty obvious why this cannot work, as the handler class passed to socketserver.ThreadingTCPServer , though it inherits from socketserver.StreamRequestHandler, overrides its __init__ method to take (rfile, wfile), which has nothing to do with the arguments that will be passed to it when being constructed.

I'm really confused by how this is set up, has this ever worked ??

gatesn:
Unclear if it has... I'm also not sure if there is much need for supporting it. Ideally, the workspace exists on the same machine as the language server so in most cases JSONRPC over stdio is sufficient. 

Either we remove it completely, or we stop trying to use the same handler class for a socket server and a stdio server. Happy to take a look here at some point.

MathieuDuponchelle:
Think I'll just fix that, I'd much rather not communicate through stdio tbh :) Do you have any idea whether other language server implementations work through stdio only ?

gatesn:
I don't know of any that work exclusively through stdio but all that I have seen support both stdio and tcp:

https://github.com/sourcegraph/go-langserver/blob/master/main.go#L17
https://github.com/felixfbecker/php-language-server/blob/master/bin/php-language-server.php#L96

etc.

[...]

PR #44, opened 2017-04-26
Fix tcp mode

MathieuDuponchelle:
Fixes issue #39 for both python 2 (2.7.11) and python 3 (3.4.3)

### DIFF
--- a/pyls/language_server.py
+++ b/pyls/language_server.py
@@ -29,6 +29,7 @@ def start_tcp_lang_server(bind_addr, port, handler_class):
         {'DELEGATE_CLASS': handler_class}
     )
 
+    socketserver.ThreadingTCPServer.allow_reuse_address = True
     server = socketserver.ThreadingTCPServer((bind_addr, port), wrapper_class)
     try:
         log.info("Serving %s on (%s, %s)", handler_class.__name__, bind_addr, port)
--- a/pyls/language_server.py
+++ b/pyls/language_server.py
@@ -70,9 +70,9 @@ class LanguageServer(JSONRPCServer):
 
     def m_initialize(self, **kwargs):
         log.debug("Language server intialized with %s", kwargs)
-        self.process_id = kwargs.get('processId')
-        self.workspace = Workspace(kwargs.get('rootPath'))
-        self.init_opts = kwargs.get('initializationOptions')
+        LanguageServer.process_id = kwargs.get('processId')
+        LanguageServer.workspace = Workspace(kwargs.get('rootPath'))
+        LanguageServer.init_opts = kwargs.get('initializationOptions')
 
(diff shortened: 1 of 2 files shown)


pyls/language_server.py
```diff
@@ -69,9 +70,9 @@ def publish_diagnostics(self, uri, diagnostics):
 
     def m_initialize(self, **kwargs):
         log.debug("Language server intialized with %s", kwargs)
-        self.process_id = kwargs.get('processId')
-        self.workspace = Workspace(kwargs.get('rootPath'))
-        self.init_opts = kwargs.get('initializationOptions')
+        LanguageServer.process_id = kwargs.get('processId')
```
gatesn: Why are we setting these at the class level?

pyls/server.py
```diff
@@ -60,32 +74,59 @@ def _content_length(self, line):
...
+            res = res.decode('utf-8')
+
+        return res
 
     def _write_message(self, msg):
         body = json.dumps(msg, separators=(",", ":"))
         content_length = len(body)
+
+        if isinstance(self.wfile, socket_type):
+            header = "HTTP/1.1 200 OK\r\n"
```
gatesn: This isn't HTTP, not sure we should be adding this header just so a requests-based client works

pyls/server.py
```diff
@@ -60,32 +74,59 @@ def _content_length(self, line):
...
         response = (
+            "{}"
             "Content-Length: {}\r\n"
             "Content-Type: application/vscode-jsonrpc; charset=utf8\r\n\r\n"
-            "{}".format(content_length, body)
+            "{}".format(header, content_length, body)
         )
+
+        if PYTHON3 and isinstance(self.wfile, socket_type):
+            response = response.encode("utf-8")
```
gatesn: Don't like if statements in general, and seems like we have a lot of them checking the socket type, if the type is bytes/unicode etc. There must be a cleaner way

pyls/server.py
```diff
@@ -32,6 +43,9 @@ def handle(self):
                 log.exception("Language server shutting down for uncaught exception")
                 break
 
+            if isinstance(self.wfile, socket_type):
+                done = True
```
gatesn: Why is done true for a socket? I think you're trying to make this support HTTP instead of the long-running TCP server. 

Check out the vscode-client project and modify src/extension.ts to use the TCP port. You will see that VSCode functions correctly sending frames over TCP

MathieuDuponchelle: I think whoever wrote this code was more than a bit confused as to how `socketserver.ThreadingTCPServer` works, and in particular the role of the handler object:

```
The request handler class must define a new handle() method, and can override any of the following methods. A new instance is created for each request.
```

PR #48, opened 2017-05-09
Support get lint config from user dir.

autozimu:


pyls/providers/lint.py
```diff
@@ -61,12 +63,10 @@ class PyCodeStyleLinter(BaseProvider):
     # Potential config files in reverse order of preference
     CONFIG_FILES = ['tox.ini', 'pep8.cfg', 'setup.cfg', 'pycodestyle.cfg']
```
autozimu: Here it seems to me the comment is not consistent with what the code does.

From the example in doc <https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read>

I am guessing settings occurred first will trump later ones. So the comment should be "in order of preferences" or the list should be reversed. I might be wrong though.

Issue #79, opened 2017-07-25
Handle urlencoded parts of document URI


tomv564:
On OS-X, when sending a hover request to _some_ locations, an exception within Jedi is caused:

```
textDocument/hover {'position': {'character': 31, 'line': 300}, 'textDocument': {'uri': 'file:///Users/tomv/Library/Application%20Support/Sublime%20Text%203/Packages/LSP/main.py'}} 
```

Useful portion of the stack here, note the urlencoded spaces:

```
File "/Users/tomv/.virtualenvs/python3/lib/python3.5/site-packages/jedi/evaluate/imports.py", line 530, in get_modules_containing_name
      for file_name in os.listdir(d):
   FileNotFoundError: [Errno 2] No such file or directory: '/Users/tomv/Library/Application%20Support/Sublime%20Text%203/Packages/LSP'
```
Adding `unquote` to every occurrence of `urlparse(...).path` should fix this issue:
```
>>> from urllib.parse import urlparse, unquote
>>> urlparse("/Users/tomv/Library/Application%20Support/Sublime%20Text%203/Packages/LSP").path
'/Users/tomv/Library/Application%20Support/Sublime%20Text%203/Packages/LSP'
>>> unquote(urlparse("/Users/tomv/Library/Application%20Support/Sublime%20Text%203/Packages/LSP").path)
'/Users/tomv/Library/Application Support/Sublime Text 3/Packages/LSP'

```

I'll supply a PR once I can verify a fix.

gatesn:
@tomv564 I just realised, my comment on your PR about unquoting first, then parsing is wrong. It allows someone to inject slashes and other reserved characters into the uri!

I'll put up a fix to revert the order, apologies.

PR #80, opened 2017-07-25
Unquote parsed urls in both Workspace and Document

tomv564:
Naïve fix for #79, where document URIs containing encoded spaces (`%20`) were not url decoded before being passed to Jedi.

Fix also applied to Workspace, as it is already written to handle URIs.

- [x] Test added for workspace and document

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -4,7 +4,7 @@ import logging
 import os
 import re
 import sys
-from urllib.parse import urlparse, urlunparse
+from urllib.parse import urlparse, urlunparse, unquote
 
 import jedi
 
@@ -25,7 +25,7 @@ class Workspace(object):
 
     def __init__(self, root, lang_server=None):
         self._url_parsed = urlparse(root)
-        self.root = self._url_parsed.path
+        self.root = unquote(self._url_parsed.path)
         self._docs = {}
         self._lang_server = lang_server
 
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @tomv564! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pyls/workspace.py
```diff
@@ -25,7 +25,7 @@ class Workspace(object):
 
     def __init__(self, root, lang_server=None):
         self._url_parsed = urlparse(root)
-        self.root = self._url_parsed.path
+        self.root = unquote(self._url_parsed.path)
```
gatesn: Would it be better to unquote, and then url parse? That way any future references to `self._url_parsed` are correct.

test/test_workspace.py
```diff
@@ -2,6 +2,7 @@
 import os
 import pytest
 from pyls import workspace
+from pyls.workspace import Workspace
```
gatesn: can you reference `workspace.Workspace` instead of adding another import.

gatesn:
Thanks very much for this! Made a couple of comments, after which it looks good to merge. Shame we can't get Circle to run for forks though.

tomv564:
Fixed both your suggestions, thanks for the review!

gatesn:
I've cloned and tested locally, all looks good! Released in 0.2.3

PR #81, opened 2017-07-26
Reverse urlencoding/unquote

gatesn:
For example, if a question mark was urlencoded then it would change the uri if it was unquoted first and then parsed.


gatesn:
Fixes #79

PR #83, opened 2017-07-26
Add signature plugin back to CORE_PLUGINS

### LINKED ISSUE #82: Signature Help support has gone missing
Likely missed during the pluggy refactoring?

Adding the signature help back in to `CORE_PLUGINS` restores this functionality.

PR coming up shortly.

tomv564:
Fixes #82 where the signature help plugin was not being registered with pluggy, resulting in empty responses.

### DIFF
--- a/pyls/plugins/__init__.py
+++ b/pyls/plugins/__init__.py
@@ -2,11 +2,11 @@
 from . import (
     completion, definition, format,
     hover, pyflakes_lint, pycodestyle_lint,
-    references, symbols
+    references, symbols, signature
 )
 
 
 CORE_PLUGINS = [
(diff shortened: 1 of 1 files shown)


pyls/plugins/__init__.py
```diff
@@ -2,11 +2,11 @@
 from . import (
     completion, definition, format,
     hover, pyflakes_lint, pycodestyle_lint,
-    references, symbols
+    references, symbols, signature
```
gatesn: Nit: can it be alphabetical? 😀

tomv564: Fixed, that was so close already :)
About building PRs from forks, is it a fixable problem? Circle's documentation mentioned the risk of leaking secrets?

gatesn (approved):

Issue #56, opened 2017-06-22
Update to LSP 3.0


gatesn:
* Use rootUri in favour of rootPath

tomv564:
I created a PR for this as I'd like to drop support for `rootPath` in my LSP client.

The other 3.0 change is versioned changes for WorkspaceEdit, but if I understand this is optional and only if the client supports it. (see
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspaceedit)

gatesn:
Once again thanks for this! I will review tomorrow and hopefully cut a release. Out of interest, what client are you working on?

________________________________

tomv564:
No problem, and thanks for this server! 
I'm working on a client for Sublime Text at https://github.com/tomv564/LSP.

gatesn:
It seems we now support LSP 3.0 given the rootUri support. Closing this ticket out since the rest of 3.0 changes are optional additions to the protocol.

PR #84, opened 2017-07-26
Support rootUri (LSP 3.0) with rootPath fallback

tomv564:
This makes rootUri the default format for the workspace directory, where rootPath is accepted and converted to a url in the initialize handler.

As discussed in #56, this is the major change from protocol version 2 to 3.

### DIFF
--- a/pyls/config.py
+++ b/pyls/config.py
@@ -3,6 +3,7 @@ from configparser import RawConfigParser
 import logging
 import os
 import pluggy
+from urllib.parse import urlparse, unquote
 
 from . import hookspecs, PYLS
 
@@ -11,8 +12,8 @@ log = logging.getLogger(__name__)
 
 class Config(object):
 
-    def __init__(self, root_path, init_opts):
-        self._root_path = root_path
+    def __init__(self, root_uri, init_opts):
+        self._root_path = unquote(urlparse(root_uri).path)
         self._init_opts = init_opts
 
         self._pm = pluggy.PluginManager(PYLS)
--- a/pyls/config.py
+++ b/pyls/config.py
@@ -13,7 +13,7 @@ log = logging.getLogger(__name__)
 class Config(object):
 
     def __init__(self, root_uri, init_opts):
-        self._root_path = unquote(urlparse(root_uri).path)
+        self._root_uri = root_uri
         self._init_opts = init_opts
 
         self._pm = pluggy.PluginManager(PYLS)
@@ -31,11 +31,12 @@ class Config(object):
         return self._init_opts
 
     @property
-    def root_path(self):
-        return self._root_path
+    def root_uri(self):
+        return self._root_uri
 
     def find_parents(self, path, names):
-        return find_parents(self.root_path, path, names)
+        root_path = unquote(urlparse(self._root_uri).path)
+        return find_parents(root_path, path, names)
 
 
 def build_config(key, config_files):
--- a/pyls/language_server.py
+++ b/pyls/language_server.py
@@ -89,10 +89,10 @@ class LanguageServer(MethodJSONRPCServer):
     def m_initialize(self, **kwargs):
         log.debug("Language server intialized with %s", kwargs)
         if 'rootUri' in kwargs:
-            self.root_uri = kwargs.get('rootUri')
+            self.root_uri = kwargs['rootUri']
         elif 'rootPath' in kwargs:
-            root_path = kwargs.get('rootPath')
-            self.root_uri = urljoin(u'file:', pathname2url(root_path))
+            root_path = kwargs['rootPath']
+            self.root_uri = urljoin(u'file://', pathname2url(root_path))
         else:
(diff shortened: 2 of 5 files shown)


pyls/language_server.py
```diff
@@ -75,22 +77,28 @@ class LanguageServer(MethodJSONRPCServer):
...
 
-    def initialize(self, root_path, init_opts, process_id):
+    def initialize(self, root_uri, init_opts, process_id):
         pass
 
     def m_initialize(self, **kwargs):
         log.debug("Language server intialized with %s", kwargs)
-        self.root_path = kwargs.get('rootPath')
+        if 'rootUri' in kwargs:
+            self.root_uri = kwargs.get('rootUri')
```
gatesn: Use `kwargs['rooturi']` instead of `get` since you've already checked existence.

pyls/language_server.py
```diff
@@ -75,22 +77,28 @@ class LanguageServer(MethodJSONRPCServer):
...
+        elif 'rootPath' in kwargs:
+            root_path = kwargs.get('rootPath')
+            self.root_uri = urljoin(u'file:', pathname2url(root_path))
+        else:
+            self.root_uri = ''
```
gatesn: Does this need to be a unicode literal? I'm not actually sure... Equally, what do we get from the kwargs?

pyls/python_ls.py
```diff
@@ -39,8 +40,10 @@ def capabilities(self):
             'textDocumentSync': lsp.TextDocumentSyncKind.INCREMENTAL
         }
 
-    def initialize(self, root_path, init_opts, _process_id):
-        self.workspace = Workspace(root_path, lang_server=self)
+    def initialize(self, root_uri, init_opts, _process_id):
+        self.workspace = Workspace(root_uri, lang_server=self)
+
+        root_path = unquote(urlparse(root_uri).path)
         self.config = config.Config(root_path, init_opts)
```
gatesn: Let's refactor the Config object to use URIs as well, should only be one place we do the conversion

test/fixtures.py
```diff
@@ -5,8 +5,13 @@
 from pyls.python_ls import PythonLanguageServer
 from pyls.workspace import Workspace
 from io import StringIO
+from urllib.parse import urljoin
+from urllib.request import pathname2url
 
 
+def path2uri(path):
+    return urljoin(u"file://", pathname2url(path))
```
gatesn, line 13: You use a double slash here, but a single slash in the workspace?

gatesn (changes requested):

pyls/language_server.py
tomv564: I have to admit ignorance on what kwargs or pathname2url returns here. Not specifying unicode literal works in python3, but in python2 I see the following when running the tests with tox
```
self.root_uri = urljoin('file://', pathname2url(root_path))
  File "/Users/tomv/Projects/palantir/python-language-server/.tox/py27/lib/python2.7/site-packages/future/backports/urllib/parse.py", line 418, in urljoin
    base, url, _coerce_result = _coerce_args(base, url)
  File "/Users/tomv/Projects/palantir/python-language-server/.tox/py27/lib/python2.7/site-packages/future/backports/urllib/parse.py", line 115, in _coerce_args
    raise TypeError("Cannot mix str and non-str arguments")
```

test/fixtures.py
tomv564: Added slashes to the language_server case for consistency

pyls/python_ls.py
tomv564: Do you mean Config should accept `root_uri` and convert it to `self.root_path`?

I agree we should convert uri -> path only once, but thought Workspace is a good place to do it as it receives root & document URIs and contains filesystem logic. 

Related: Workspace has an `is_local` method which is only used in a test. If we remove this function, Workspace also has no need for its `_url_parsed` field.

gatesn: Yeah, we kind of lost support for non-local workspaces in the plug upgrade. happy to add it back one day. 

Since we're pre-1.0 I wouldn't worry too much about API breaks. So I'd just replace the Config field with root_uri (not path), have Workspace also take a root_uri, and then parse the path once in the initialisation opts

tomv564: Thanks for the clarification, got a chance to implement your suggestions just now!

pyls/config.py
```diff
@@ -11,8 +12,8 @@
 
 class Config(object):
 
-    def __init__(self, root_path, init_opts):
-        self._root_path = root_path
+    def __init__(self, root_uri, init_opts):
+        self._root_path = unquote(urlparse(root_uri).path)
```
gatesn: Sorry, I meant to just store a `self._root_uri` and expose a `root_uri` property. Breaking the API of the Config object, but it doesn't really matter.

gatesn (approved):
Looks good, thanks!

Issue #85, opened 2017-08-05
textDocument/definition returns wrong path on windows


prabirshrestha:
Notice the wrong path where d drive occurs twice. ```{"uri":"file:///D:\\D:\\tmp\\python-helloworld\\helloworld.py"...```. `textDocument/documentSymbol` seems to work.

The log is from `vim-lsp`.

```vim
8/5/2017 3:59:30 PM:["lsp#register_server","server registered","pyls"]
8/5/2017 3:59:30 PM:["lsp#register_server","server registered","rls"]
8/5/2017 3:59:30 PM:["s:on_text_document_did_open()",1]
8/5/2017 3:59:30 PM:[{"response":{"data":{"__data__":"vim-lsp","lps_id":1,"server_name":"pyls"},"message":"started lsp server successfully"}}]
8/5/2017 3:59:30 PM:["--->",1,"pyls",{"method":"initialize","params":{"rootUri":"file:///D:/tmp/python-helloworld","capabilities":{},"rootPath":"file:///D:/tmp/python-helloworld"}}]
8/5/2017 3:59:34 PM:["<---",1,"pyls",{"response":{"id":1,"jsonrpc":"2.0","result":{"capabilities":{"codeLensProvider":{"resolveProvider":false},"executeCommandProvider":{"commands":[]},"hoverProvider":true,"documentFormattingProvider":true,"signatureHelpProvider":{"triggerCharacters":["(",","]},"documentSymbolProvider":true,"definitionProvider":true,"completionProvider":{"resolveProvider":false,"triggerCharacters":["."]},"codeActionProvider":true,"referencesProvider":true,"documentRangeFormattingProvider":true,"textDocumentSync":2}}},"request":{"method":"initialize","jsonrpc":"2.0","id":1,"params":{"rootUri":"file:///D:/tmp/python-helloworld","capabilities":{},"rootPath":"file:///D:/tmp/python-helloworld"}}}]
8/5/2017 3:59:34 PM:["--->",1,"pyls",{"method":"textDocument/didOpen","params":{"textDocument":{"uri":"file:///D:/tmp/python-helloworld/helloworld.py","version":1,"languageId":"python","text":"#!/usr/bin/env python\nclass MyClass:\n    variable = \"blah\"\n\n    def function(self):\n        print(\"This is a message inside the class.\")\n\nmyobjectx = MyClass()\n\nmyobjectx.variable"}}}]
[...]

PR #87, opened 2017-08-07
Handle windows paths in workspace.get_uri_like + test

tomv564:
Fix for #85 where windows paths were not prefixed and converted to forward slashes

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -4,7 +4,7 @@ import logging
 import os
 import re
 import sys
-from urllib.parse import urlparse, urlunparse, unquote
+from urllib.parse import urlparse, urlunparse, quote, unquote
 
 import jedi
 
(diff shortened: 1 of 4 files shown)


gatesn (commented):
Good catches here. One thing I'd say, is do we need to url quote the path we use in the uri_like function? I know it's not part of this Pr but may as well slip it in

tomv564:
Sounds like a good idea, I'll add quoting. 
It will try to (unnecessarily) encode the `:` in `/D:/myfile` which I'd like to avoid.
I couldn't find any reason why module_path is sometimes None, please let me know if my assumption was correct!

pyls/workspace.py
```diff
@@ -181,5 +181,10 @@ def get_uri_like(doc_uri, path):
     unicode objects.
     """
     parts = list(urlparse(doc_uri))
+    if path[0] != '/':  # fix path for windows
```
gatesn: Can we add `and ':' in path` just to make sure we don't throw on the next line

gatesn (approved):
LGTM

Issue #96, opened 2017-08-19
Frequently hanging on Windows


deathaxe:
After fixing crashes (#95) caused by invalid path translations, Sublime Text still freezes frequently, while editing. In such situations python.exe which runs the pyls seems to run in an endless loop consuming 100% CPU. Killing the process causes ST to work normally again.

Something seems to cause a deadlock which blocks the whole stdin/stdout communication.

The logfile does not contain any errors/warnings.

### Setup
- Windows 10 1703 x64
- Sublime Text 3 4142 with LSP package
- python-language-server 0.4.1
- both connected via stdin/stdout interface
- python 3.6.2 x64

gatesn:
My suspicion would be that on Windows, this isn't the correct base case to break out of the loop: https://github.com/palantir/python-language-server/blob/develop/pyls/config.py#L79

deathaxe:
Seems to be the point. Investigating ...

PR #95, opened 2017-08-18
Fix translation between uri and local file path on Windows (#94)

deathaxe:

PR #99, opened 2017-08-20
Better support for Windows

gatesn:


deathaxe:
Had a very very short view on it. So nothing final, but as python starts eating 100% CPU again I saw you didn't merge https://github.com/palantir/python-language-server/pull/95/commits/74fe7be70dc5d06a94ae04b3a65818395bbbce05 right now. It is required. Otherwise we'll end up in #96 - the endless loop which freezes the language-server

pyls/uris.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2017 Palantir Technologies, Inc.
+"""A collection of URI utilities with logic built on the VSCode URI library.
+
+https://github.com/Microsoft/vscode-uri/blob/e59cab84f5df6265aed18ae5f43552d3eef13bb9/lib/index.ts
+"""
+import re
+from urllib import parse
+from pyls import IS_WIN
+
+RE_DRIVE_LETTER_PATH = re.compile(r'^\/[a-zA-z]:')
```
deathaxe, line 10: Better: r'^\/[a-zA-Z]:'

A-z may catch unwanted characters.

pyls/uris.py
```diff
@@ -0,0 +1,129 @@
...
+    Will handle UNC paths and normalize windows drive letters to lower-case. Also
+    uses the platform specific path separator. Will *not* validate the path for
+    invalid characters and semantics. Will *not* look at the scheme of this URI.
+    """
+    # scheme://netloc/path;parameters?query#fragment
+    scheme, netloc, path, _params, _query, _fragment = urlparse(uri)
+
+    if netloc and path and scheme == 'file':
+        # unc path: file://shares/c$/far/boo
+        value = "//{netloc}{path}".format(netloc=netloc, path=path)
```
deathaxe, line 58: Why that long? Is `"//{}{}".format(netloc, path)` not obvious enough here?

pyls/workspace.py
```diff
@@ -24,19 +23,28 @@ class Workspace(object):
     M_SHOW_MESSAGE = 'window/showMessage'
 
     def __init__(self, root_uri, lang_server=None):
-        self._url_parsed = urlparse(root_uri)
-        self.root = unquote(self._url_parsed.path)
+        self._root_uri = root_uri
+        self._root_uri_scheme = uris.urlparse(self._root_uri)[0]
+        self._root_path = uris.to_fs_path(self._root_uri)
```
deathaxe, line 28: I wonder if there is a way to avoid calling urlparse twice here. Even though it might not be a big deal it is a waste of resources, calling `uris.urlparse()` to convert the whole path and throw away everything but the scheme and then calling `uris.urlparse()` again in `uris.to_fs_path()`.

pyls/workspace.py
```diff
@@ -24,19 +23,28 @@ class Workspace(object):
...
+    def root_path(self):
+        return self._root_path
+
+    @property
+    def root_uri(self):
+        return self._root_uri
+
     def is_local(self):
-        return (self._url_parsed.scheme == '' or self._url_parsed.scheme == 'file') and os.path.exists(self.root)
+        return (self._root_uri_scheme == '' or self._root_uri_scheme == 'file') and os.path.exists(self._root_path)
```
deathaxe, line 41: This function is not used at all. Are there plans to do so in the future?

pyls/workspace.py
```diff
@@ -66,21 +74,21 @@ def syspath_for_path(self, path):
...
         """
-        files = config.find_parents(self.root, path, ['setup.py']) or []
+        files = config.find_parents(self._root_path, path, ['setup.py']) or []
         path = [os.path.dirname(setup_py) for setup_py in files]
         path.extend(sys.path)
         return path
 
     def is_in_workspace(self, path):
-        return not self.root or os.path.commonprefix((self.root, path))
+        return not self._root_path or os.path.commonprefix((self._root_path, path))
```
deathaxe, line 83: This function is not used, too? Same question as with is_local()

pyls/uris.py
gatesn: Ah that's interesting, looks like it's the same in the VSCode package: https://github.com/Microsoft/vscode-uri/blob/e59cab84f5df6265aed18ae5f43552d3eef13bb9/lib/index.ts#L56

gatesn: Created an issue here: https://github.com/Microsoft/vscode/issues/32961

pyls/uris.py
gatesn: Yup, fair point

pyls/workspace.py
gatesn: Given this is done once per language server I'm not too fussed. If it was done once per RPC, it might be a slightly bigger issue.

pyls/workspace.py
gatesn: I think it was used previously... we should add back support for non-local URIs and stop assuming things exist on disk. Let me create issues for these things.

pyls/workspace.py
gatesn: Removed this one, it used to be used by find_parents, but now that checks itself without needing a workspace.

gatesn (commented):
Thanks for this review!! I've address your comments in #103

brandondube:
Any updates on this?  LSP/pyls have become so unstable for me lately that I have moved to VSCode in the interim even though its completions are worse and I don't have a proper docstring generator.

deathaxe:
This branch is merged.

The frequent deadlocks of LSP are not related to this PR. The issue is discussed here: https://github.com/tomv564/LSP/issues/287

As I didn't track the development for a while now, I am out of info about the possible source of issue. 

Using the logging features of LSP and pyls doesn't help as the relevant entries are not flushed to disk, if pyls freezes. After killing pyls those are lost, but ST continues working.

Very annoying and nasty bug as it happens each 10 minutes or so.

EDIT: As VS Code is too bloated I switched back to _Python - Autocompletion_ in combination with _SublimeLinter_. Works great with nearly the same amount of features but without the annoying .rope stuff which pollutes any project with its stupid config files.

PR #111, opened 2017-09-03
Respect yapf style config

### LINKED ISSUE #105: Support .style.yapf
I tried using this to autoformat code, and it seemed to ignore the settings in my .style.yapf in the cwd.

gatesn:
Fixes #105


gatesn:
@brennie does this work for you?

perrinjerome:
This did not work for me :) 

`document.filename` is not the full path of the file, just the base name. `document.path` should be used instead if I understood correctly. Can you please take a look at https://github.com/palantir/python-language-server/pull/473 ?

PR #121, opened 2017-09-04
Pylama experiments

### LINKED ISSUE #104: pylama
Great work so far!

I have two small suggestion - for linter, would you consider changing pydocstyle+pyflakes to pylama? Pylama covers pretty much all the essential linters in one: pep8 (now pycodestyle), pydocstyle, pyflakes, pylint and McCabe. Easier than adding a linter for each. The recently added mypy is a nice optional addition.

I think Jedi + pylama + YAPF + isort would be a a very comprehensive combination.

gatesn:
Fixes #104 

@JGH1000 so it seems pretty easy to add PyLama as a linter to pyls. The issue I have, is that Pylama doesn't respect configuration options in the same way that the original linters (pycodestyle, pyflakes, etc.) do.

What this means is that you have to be aware that you're using pylama. 

Therefore I think it's worth adding as a linter, but defaults to disabled. If you wish to use PyLama then you can enable it and disable the other linters.

I can create tickets for the other linters that PyLama uses so we can support their messages, I just don't like the way PyLama's configuration works.


ghost:
Thanks for a quick inclusion! Having several linters included with an easy option to turn on/off is a perfect solution. Especially for people using different combination of them.

gatesn:
@JGH1000 Another reason is that PyLama doesn't seem to be written for inclusion into other programs. You can see that as well as the logging (which I've included a module-level statement to disable), the tests all fail once pylama is imported! So it must be doing other bad things at the module level.

ghost:
@gatesn If it is really that tricky perhaps it is better to just provide all the packages within pylama directly - pylama is simply a wrapper. You already have pyflakes and pycodestyle so the big missing ones are pydocstyle, pylint and mccabe. Probably a more elegant solution but I realise it is more work on your part.

Anyway, thanks for expanding lining capabilities! Pretty sure Atom will move to this solution as well in the near future.

gatesn:
@JGH1000 Couple of choices here, I could merge this and leave it defaulting to disabled. Or if you want PyLama for any reason (over just the collection of other linters, e.g. because all your configs are for pylama), then it might be worth creating a separate package similar to https://github.com/tomv564/pyls-mypy

ghost:
@gatesn I am thinking of what would be the best for users - probably adding pydocstyle, pylint and mycabe as standalone linters would be better than pylama wrapper since it gives people more flexibility. I am absolutely fine changing my configs if you decide to go this route.

Thanks a lot for your effort!

PR #140, opened 2017-09-22
Jedi symbols: Add all_scopes config option

lgeiger:
This PR adds a new config option to allow users to configure if symbols from all scopes are shown or not.

This was originally brought up on the `ide-python` package for Atom: https://github.com/lgeiger/ide-python/issues/11 

## `all_scopes=false` (default / old behavior)
all_scopes=false (default)

## `all_scopes=true`
all_scopes=true

**Note:** I'm using Atom to test this and haven't tried the VSCode plugin, but judging from the existing config options this should work as expected.

### DIFF
--- a/pyls/plugins/symbols.py
+++ b/pyls/plugins/symbols.py
@@ -8,7 +8,7 @@ log = logging.getLogger(__name__)
 
 @hookimpl
 def pyls_document_symbols(config, document):
-    all_scopes = config.plugin_settings('jedi_symbols').get('all_scopes', False)
+    all_scopes = config.plugin_settings('jedi_symbols').get('all_scopes', True)
     definitions = document.jedi_names(all_scopes=all_scopes)
     return [{
         'name': d.name,
--- a/pyls/plugins/symbols.py
+++ b/pyls/plugins/symbols.py
@@ -7,8 +7,9 @@ log = logging.getLogger(__name__)
 
 
 @hookimpl
-def pyls_document_symbols(document):
-    definitions = document.jedi_names()
+def pyls_document_symbols(config, document):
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @lgeiger! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn:
I’m not opposed to turning this on by default if you think it makes sense?

Wonder if we can get a better list of symbols though, as this seems to include method parameters.

gatesn:
@lgeiger thanks very much for this! Looks good. If you could change the default to true (not too fussed about changing existing behaviour since we're still pre-1.x), then I'd be happy to merge this.

lgeiger:
Sorry for the late reply.

> Wonder if we can get a better list of symbols though, as this seems to include method parameters.

Unfortunately this seams to be jedi's fault. When calling `jedi_names` with `all_scopes=false` on:
```python
def test(param1, param2):
    pass
```
It also returns both `test`, `param1`  and `param2`.

> @lgeiger thanks very much for this! Looks good. If you could change the default to true (not too fussed about changing existing behaviour since we're still pre-1.x), then I'd be happy to merge this.

👍 I'll change that.

I've been experimenting with properly structuring the symbols using [`containerName`](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#document-symbols-request) but I'm getting strange UI behavior in conjunction with Atom's language server plugin. I'll submit another PR once I know if the bug is inside my code or comes from Atom's language server plugin.

PR #146, opened 2017-09-29
Allow configuring pycodestyle options from the PYLS config

### LINKED ISSUE #144: Custom pycodestyle configuration via xcode/Atom settings
It would be really great to configure the ignored error codes and the maximum line length of `pycodestyle` via the settings if no configuration file is found on the system.

This is the most requested issue on the Atom package for `pyls`: https://github.com/lgeiger/ide-python/issues/9

gatesn:
Fixes #144

### DIFF
--- a/pyls/plugins/pycodestyle_lint.py
+++ b/pyls/plugins/pycodestyle_lint.py
@@ -21,29 +21,29 @@ def pyls_lint(config, document):
     conf_to_use = pycodestyle_conf if pycodestyle_conf else pep8_conf
 
     conf = {k.replace("-", "_"): v for k, v in conf_to_use.items()}
-    log.debug("Got pycodestyle config: %s", conf)
 
     # Grab the pycodestyle parser and set the defaults based on the config we found
     parser = pycodestyle.get_parser()
-    par.ser.set_defaults(**conf)
+    parser.set_defaults(**conf)
 
     # Override with any options set in the language server config
-    args = []
+    argv = []
     ls_conf = config.plugin_settings('pycodestyle')
-    if 'exclude' in ls_conf:
-        args.extend(['--exclude', ','.join(ls_conf['exclude'])])
-    if 'filename' in ls_conf:
-        args.extend(['--filename', ','.join(ls_conf['filename'])])
-    if 'select' in ls_conf:
-        args.extend(['--select', ','.join(ls_conf['select'])])
-    if 'ignore' in ls_conf:
-        args.extend(['--ignore', ','.join(ls_conf['ignore'])])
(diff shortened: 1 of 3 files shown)


lgeiger:
Thanks a lot for the super fast fix, this is awesome.

Though I feel like the config from files like `setup.cfg` should take precedence over the global config defined in the VSCode settings. This way people can very easily switch between projects that don't have specific pycodestyle configs (and therefor use the defaults set in the VSCode config) and still have projects with custom configs. What do you think?

gatesn:
Are there any precedents in this space? I’m trying to draw parallels with things like IntelliJ checkstyle inspections etc.

The config sources are:

* files in a user’s home directory (user configs)
* files in the project (project configs)
* within the file (e.g. # noqa)
* IDE settings (the ones in question)

I’m going to take linting as the example here. Obviously “within the file” should win always, so let’s dismiss those settings.

My thinking is that project configs are configured to be verified at CI/build time. They basically say, “I don’t care who works on the project, they need to conform to this style guide”.

User configs (in their home directory) are there for when project configs don’t exist. You’re working on a project that doesn’t have any style preference explicitly checked in, so you’re going to apply your own preferences for your work only.

So project configs override user configs when running the linter.

IDE settings then exist to customise the user’s editing experience. In applying these settings they are choosing to ignore what the project says, and go with their own style. So they should override the project settings.

* If you want to default project configurations, you should do so in your user configs.
* if you want those configs verified at build time, they should be checked in to the project
* if you want to ignore the project settings and apply your own, then you configure your editor.

At least that was my train of thought implementing this. I could probably argue for the other way as well though! So keen to hear your thoughts.


________________________________

lgeiger:
Thanks a lot for the thorough explanation of your train of thought!

I'm seeing the IDE settings as a replacement for the user configs (files in a user’s home directory) in order to give not very technical people an easy and fast way of configuring the linting behavior for projects with no custom style guide defined in a project config.

Generally speaking I don't think it's a particular good thing to give people the possibility to overwrite project configs. In some cases it can eliminate the benefit of having a project config in, being that it removes the nitpicking about code style in code reviews.

> Are there any precedents in this space?

[linter-flake8](https://github.com/AtomLinter/linter-flake8#project-configuration-files) and [prettier](https://github.com/prettier/prettier-atom#usage) both use the IDE configuration only as a fallback when no project configs are found.

lgeiger:
Would you be interested in a PR implementing the behavior above?

gatesn:
@lgeiger could you review https://github.com/palantir/python-language-server/pull/156 please?

PR #155, opened 2017-10-03
Decrease module completion time by using both jedi and rope

### LINKED ISSUE #108: Support rename
Renaming things without regex is nice.

### LINKED ISSUE #147: Provide access to rope as well as Jedi
We're currently developing an LSP extension for Spyder https://github.com/spyder-ide/spyder/pull/4751 and we're using your server to test all the LSP features, and eventually we'll be using it by default. 

At the moment we haven't got much trouble migrating our own implemented functionalities, however we have noticed that some ``textDocument/completion`` requests take too long to be replied by the server, especially if they are related to modules such as numpy. On such cases we've experienced about 3-5 delay seconds.

Right now on Spyder we've been using both Jedi and Rope to handle our autocompletion calls, as Rope is faster than Jedi in resolving autocompletions for large/compiled modules, we achieve autocompletion in less than a second, which is one of the features most beloved by the Spyder user community.

We would like to contribute to improve the autocompletion performance of the server, we propose to use both Jedi and Rope to handle LSP completion requests, the idea is to define a race condition between both packages, by returning the results of the package that completes the autocompletion task first.

We would like to hear your opinion on this solution and discuss other alternatives, as Spyder LSP migration is imminent.

### LINKED ISSUE #148: Faster completions with rope
c.f. #147

andfoy:
Fixes #147 
Fixes #148 
Fixes #108


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @andfoy! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pyls/workspace.py
```diff
@@ -16,11 +20,59 @@
...
+            submodules.append(sm_name)
+    except ImportError:
+        return []
+    except Exception:
+        return [mod]
+    return submodules
+
+
+def get_preferred_submodules():
+    mods = ['numpy', 'scipy', 'sympy', 'pandas',
```
andfoy, line 43: @gatesn Do you agree to preload all these modules?

gatesn: Supposing we do preload all of these modules. What's the Jedi performance like after the initial import? I've never run into an issue with Jedi after the first import.

andfoy: Well, the preloading only affects Rope. With respect to jedi, I found that it is possible also to preload modules, I'll make some comparisons

andfoy: Maybe @ccordoba12 has more information about which features Rope has that Jedi doesn't

pyls/plugins/completion.py
```diff
@@ -3,28 +3,89 @@
 from pyls.lsp import CompletionItemKind
 from pyls import hookimpl
 
+from threading import Thread, Lock
+from rope.contrib.codeassist import code_assist, sorted_proposals
+
 log = logging.getLogger(__name__)
 
 
+class CompletionThread(Thread):
```
gatesn: It might also be nice to generalise this somehow. Ideally, we could call any of the pluggy hooks in a "race" mode where the first non-None response wins.

andfoy: @gatesn Would you like to extract these kind of methods to anywhere else?

pyls/workspace.py
andfoy: @gatesn I've found that rope is only faster on the preloaded modules than Jedi

andfoy:
@gatesn @ccordoba12 Numpy autocompletion time has been reduced significantly!  
rope-lsp

pyls/workspace.py
ccordoba12: Yep, Rope is *significantly* faster to complete compiled modules than Jedi.

@andfoy, could we expose this value somehow, in case we want to add more modules to it in Spyder?

setup.py
```diff
@@ -40,6 +40,7 @@
         'pycodestyle',
         'pydocstyle',
         'pyflakes',
+        'rope',
```
ccordoba12: Please change this to

    rope>=0.10.5

which supports Python 2 and 3 using the same source base.

vscode-client/src/extension.ts.orig
```diff
@@ -0,0 +1,54 @@
+/* --------------------------------------------------------------------------------------------
```
ccordoba12: Please remove this file.

ccordoba12:
@andfoy, this is looking really great! Thanks for implementing it in a really short time!!

@gatesn, what do you think so far?

pyls/workspace.py
andfoy: @ccordoba12, unfortunately, under the LSP specs, we cannot do such thing

gatesn: @andfoy can't we take them as configuration options? https://github.com/palantir/python-language-server/blob/develop/vscode-client/package.json#L22

pyls/workspace.py
```diff
@@ -29,6 +81,17 @@ def __init__(self, root_uri, lang_server=None):
         self._docs = {}
         self._lang_server = lang_server
 
+        # Whilst incubating, keep private
+        self.__rope = Project(self._root_path)
+        self.__rope.prefs.set('extension_modules', self.PRELOADED_MODULES)
+        # jedi.api.preload_module(*self.PRELOADED_MODULES)
```
gatesn: What did you discover / decide with this?

pyls/workspace.py
```diff
@@ -29,6 +81,17 @@ def __init__(self, root_uri, lang_server=None):
...
+
+    @property
+    def _rope(self):
+        # TODO: we could keep track of dirty files and validate only those
+        self.__rope.validate()
```
gatesn, line 91: Does this actually work? I never tested it?

pyls/plugins/completion.py
gatesn: Yes please. I think the best way to structure this is to rename this to `jedi_completion.py` and create a new `rope_completion.py`. Then in `python_ls.py` you can grab the hooks from pluggy and run them through this "racer" that runs all hooks and returns the first result.

Reasoning is:

* There may be more things in the future that we want to race like this.
* This helps us step towards the LSP streaming extension (return one set of results, then update them some time later)
* People can explicitly choose to disable/enable rope or jedi if they have any problems

pyls/workspace.py
andfoy: @gatesn Yes we can, but those configuration settings are only used if LSP is launched on VSCode, isn't it?

pyls/workspace.py
andfoy: I will take a look at it, however it's not generating any error

pyls/workspace.py
andfoy: Rope autocompletion of preloaded modules is much faster than Jedi, also I found that by using the Jedi preloading functions, the server gets blocked.

andfoy:
@gatesn I've just extracted the parallel logic to ``python_ls.py``, please tell me if it's ok for you. If it is, I'll write some tests

pyls/workspace.py
gatesn: I think most clients have some interpretation of these configuration options.

pyls/workspace.py
gatesn: The problems I ran into were performing a workspace rename, then undoing, then trying to do it again. Haven't tested this new implementation yet though!

pyls/hookspecs.py
```diff
@@ -23,7 +23,12 @@ def pyls_commands(config, workspace):
 
 
 @hookspec
-def pyls_completions(config, workspace, document, position):
+def pyls_rope_completions(config, workspace, document, position):
```
gatesn: We should keep `pyls_completions` and just have the rope and jedi plugins implement them separately.

andfoy: Well, by doing this, you cannot control each implementation separately, instead both are called simultaneously

gatesn: That's what I meant by moving the race logic into the pluggy layer. So for any pluggy hook we can race the results. I can take a look to see if its possible

gatesn:
@andfoy started playing around, but noticed completions can be quite slow: 
slowcompletions

andfoy:
@gatesn I'll just correct the merge conflicting files, and we should be ready to go, isn't it?

test/plugins/test_completion.py
```diff
@@ -1,7 +1,15 @@
 # Copyright 2017 Palantir Technologies, Inc.
+import os
 from pyls import uris
-from pyls.plugins.completion import pyls_completions
-from pyls.workspace import Document
+import os.path as osp
+from rope.base import libutils
+from rope.base.project import Project
+from pyls.workspace import Document, get_preferred_submodules
+from pyls.plugins.jedi_completion import pyls_jedi_completions
```
gatesn: Think you need `from pyls.plugins.jedi_completion import pyls_completions as pyls_jedi_completions `

andfoy: That's right

gatesn:
@andfoy thanks very much for this! Played around and all seems performant and stable. Happy to merge after you fix conflicts and tests.

gatesn (approved):

PR #156, opened 2017-10-03
Use better overrides order

gatesn:
Allow project configuration to override IDE configuration.

c.f. #146

### DIFF
--- a/pyls/_utils.py
+++ b/pyls/_utils.py
@@ -1,6 +1,10 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import functools
+import re
 import threading
+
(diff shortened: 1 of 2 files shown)


lgeiger:
Wow thanks for the fast response 🎉 

I tried it out with Atom, but unfortunately I can't get the `ignore` setting to work. `max-line-length`, `select` and `hang-closing` are working fine though.

I don't know where the problem is. Is it working in VSCode?

gatesn:
Oh I do, the settings are a list, but pycodestyle wants them as a comma separated string. Whoops!

________________________________

PR #157, opened 2017-10-03
Use import magic for auto-imports

gatesn:
@tomv564 would you mind testing this out? I'm not 100% that importmagic is stable enough, but can give it a go.


gatesn:
Actually, I think I'd rather base this functionality on isort instead. Going to close this out for now.

gmoutso:
Is auto import implemented?
Edit: I found [a pull request](https://github.com/palantir/python-language-server/pull/685). Looking forward to it getting accepted.

PR #164, opened 2017-10-20
pycodestyle config: convert arrays to comma seperated strings

lgeiger:
This fixes https://github.com/palantir/python-language-server/pull/156#issuecomment-334129877 since pycodestyle expects the config values as a comma separated string.

Unfortunately ignoring a specific error code still doesn't work correctly, all errors are ignored instead. The same happens for warnings too. Any idea why this is happening?

Fixes #156


gatesn:
Thanks for this! Is there a separate ticket open for the other issue? I can look into some time this week probably

lgeiger:
> I can look into some time this week probably

Thanks a lot, that would be great 👍 

> Is there a separate ticket open for the other issue?

 I opened #165 which adds a failing test case to illustrate the issue.

Issue #149, opened 2017-10-01
Reliable content for CompletionItem details field


tomv564:
In https://github.com/tomv564/LSP/pull/136 a user has decided they do not want to see the CompletionItem `details` because the content is unpredictable and most of the time repeats the `label` field's content:

![screenshot](https://user-images.githubusercontent.com/16542113/31048209-1179a0c2-a619-11e7-9a8e-7e53bacc39cc.png)

PyCharm elects to show the completion's module name in the 2nd column:

<img width="457" alt="screen shot 2017-10-01 at 09 47 06" src="https://user-images.githubusercontent.com/4395832/31052753-edc10fba-a68d-11e7-82a7-09eb7b6c3e9c.png">

The protocol suggests the following about the content of the `detail` field:
```typescript
/**
 * A human-readable string with additional information
 * about this item, like type or symbol information.
 */
detail?: string;
```

The Sublime LSP package will likely provide an option to always display "kind" instead of "detail". 

Jedi's CompletionItem exposes a `module_name` field from BaseDefinition.
Some `python-language-server` users may prefer this content in the "detail" field. 
What do you think?

tomv564:
`module_name` was a bad guess, what I really wanted was the **parent** of the definition.

Using `parent().full_name` gets pretty close:

<img width="320" alt="screen shot 2017-10-01 at 20 32 14" src="https://user-images.githubusercontent.com/4395832/31057712-ac475a80-a6e7-11e7-893b-328067dbf37e.png">

We can also add parameters to the completion labels of functions and methods:

<img width="658" alt="screen shot 2017-10-01 at 20 30 42" src="https://user-images.githubusercontent.com/4395832/31057691-7230942e-a6e7-11e7-877f-bcd597dd8495.png">

That said, we're not trying to turn the language server into a PyCharm engine, so I'd love to hear what other users think. Iterating through Jedi definition's parameters probably has a cost as well.

gatesn:
Agree the current details aren't great. I like the `parent().full_name`, wonder if there's anything cheaper as you say.

Jedi Vim uses the type + full module path:

![screenshot_complete](https://user-images.githubusercontent.com/1928052/31143858-a997aaa2-a876-11e7-9fa2-612f4b1c9cdf.png)

VSCode uses a combination of signature and first line of docs: https://github.com/DonJayamanne/pythonVSCode/blob/master/src/client/providers/completionProvider.ts#L27

[...]

PR #170, opened 2017-10-24
Completion detail using module name, function parameters

tomv564:
Here are some explorations for #149 for testing / discussion. Likely to conflict with the Rope integration, but this is far from decided anyways.

screen shot 2017-10-01 at 20 30 42

Performance is okay for me in a small codebase, anyone working with a large code base who can do a quick comparison?
References to Jedi implementations:

Signature params: https://github.com/davidhalter/jedi/blob/752b7d8d495689a38076f20475a870f8aa139a28/jedi/api/classes.py#L314

Full_name:
https://github.com/davidhalter/jedi/blob/752b7d8d495689a38076f20475a870f8aa139a28/jedi/api/classes.py#L265

### DIFF
--- a/pyls/plugins/completion.py
+++ b/pyls/plugins/completion.py
@@ -10,12 +10,21 @@ log = logging.getLogger(__name__)
 def pyls_completions(document, position):
     definitions = document.jedi_script(position).completions()
     return [{
-        'label': d.name,
+        'label': _label(d),
         'kind': _kind(d),
         'detail': _detail(d),
         'documentation': d.docstring(),
-        'sortText': _sort_text(d)
+        'sortText': _sort_text(d),
+        'insertText': d.name
     } for d in definitions]
+
+
+def _label(definition):
+    if definition.type in ('function', 'method'):
+        params = ", ".join(map(lambda definition: definition.name, definition.params))
+        return "{}({})".format(definition.name, params)
+
+    return definition.name
 
 
 def _detail(definition):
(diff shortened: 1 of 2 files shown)


gatesn:
@tomv564 thanks for this, I like it! Much more useful IMO.

Happy to merge once tests pass.

gatesn:
@tomv564 think you need `read()` not `read(args)`

[edit] ahh, it's a py2 vs py3 thing

pyls/plugins/completion.py
```diff
@@ -10,14 +10,27 @@
...
         'documentation': d.docstring(),
-        'sortText': _sort_text(d)
+        'sortText': _sort_text(d),
+        'insertText': d.name
     } for d in definitions]
 
 
+def _label(definition):
+    if definition.type in ('function', 'method'):
+        params = ", ".join(map(lambda definition: definition.name, definition.params))
```
gatesn: Also, could you use a list comprehension instead of the `map` builtin.

test/plugins/test_completion.py
```diff
@@ -23,7 +23,7 @@ def test_completion():
     items = pyls_completions(doc, com_position)
 
     assert len(items) > 0
-    assert items[0]['label'] == 'read'
+    assert items[0]['label'] == 'read()'
```
gatesn: Doesn't make sense to try and deal with 2vs3 (I believe python 3 lets you "inspect" the python * and ** parameters to a function), so you may want to just pick another builtin to test against instead of `sys.stdin.read` !

tomv564: Tried `sys.exit` which also differed, for now I added the version check inline.
I can try a few more later, agree it is clumsy logic to have in the tests.

PR #175, opened 2017-10-28
Correctly structure outline view

lgeiger:
This correctly structures the outline view and hides function parameter and variable assignments inside functions:

bildschirmfoto 2017-10-28 um 15 54 22

Fixes https://github.com/atom/atom-languageclient/issues/116

### DIFF
--- a/pyls/plugins/symbols.py
+++ b/pyls/plugins/symbols.py
@@ -9,18 +9,47 @@ log = logging.getLogger(__name__)
 @hookimpl
 def pyls_document_symbols(config, document):
     all_scopes = config.plugin_settings('jedi_symbols').get('all_scopes', True)
-    definitions = document.jedi_names(all_scopes=all_scopes)
-    return [{
-        'name': d.name,
-        'kind': _kind(d),
-        'location': {'uri': document.uri, 'range': _range(d)}
-    } for d in definitions]
+    names = document.jedi_names(all_scopes=all_scopes)
+    return list(filter(None, [serialize_name(n, document.uri) for n in names]))
 
 
-def _range(d):
+def serialize_name(name, uri):
+    result = {}
+    parent = None
+
+    try:
+        # Jedi sometimes fails here.
+        parent = name.parent()
+    except:
+        pass
+
+    # - Only include global assignments in outline.
+    # - Don't include underscore assignments in outline.
+    # - Don't include function parameters in outline.
+    if (name.type == 'statement' and
+        (name.name == '_' or parent is None or
+         (parent and parent.parent()))) or name.type == 'param':
+        return None
+    else:
+        result['name'] = name.name
+
+    result['location'] = {'uri': uri, 'range': _range(name)}
+    result['kind'] = _kind(name)
+
+    # All assignments have the filename as a there toplevel parent.
+    if parent and parent.parent():
+        result['containerName'] = parent.name
+
+    return result
+
+
+def _range(name):
+    definition = name._name.tree_name.get_definition()
+    (start_line, start_column) = definition.start_pos
+    (end_line, end_column) = definition.end_pos
     return {
-        'start': {'line': d.line - 1, 'character': d.column},
-        'end': {'line': d.line - 1, 'character': d.column + len(d.name)}
+        'start': {'line': start_line - 1, 'character': start_column},
+        'end': {'line': end_line - 1, 'character': end_column}
     }
 
 
--- a/test/plugins/test_symbols.py
+++ b/test/plugins/test_symbols.py
@@ -10,11 +10,14 @@ DOC = """import sys
 a = 'hello'
 
 class B:
-    def __init__():
-        pass
+    def __init__(self):
+        x = 2
+        self.y = x
 
-def main():
-    pass
+def main(x):
+    y = 2 * x
+    return y
+
 """
 
 


pyls/plugins/symbols.py
```diff
@@ -9,18 +9,47 @@
...
+
+    try:
+        # Jedi sometimes fails here.
+        parent = name.parent()
+    except:
+        pass
+
+    # - Only include global assignments in outline.
+    # - Don't include underscore assignments in outline.
+    # - Don't include function parameters in outline.
```
gatesn, line 28: Python language server shouldn't have opinions about this. The client can decide to filter out these symbols if it wishes.

pyls/plugins/symbols.py
```diff
@@ -9,18 +9,47 @@
...
+    if (name.type == 'statement' and
+        (name.name == '_' or parent is None or
+         (parent and parent.parent()))) or name.type == 'param':
+        return None
```
gatesn, line 32: Not a fan of returning `None` to indicate something. Should refactor and use as a predicate in the filter.

pyls/plugins/symbols.py
```diff
@@ -9,18 +9,47 @@
...
+    else:
+        result['name'] = name.name
+
+    result['location'] = {'uri': uri, 'range': _range(name)}
```
gatesn, line 36: Feels weird to pass the uri in here just to construct the result. Equally calling _range from within this function doesn't feel right if the return value is being used as-is in the result.

pyls/plugins/symbols.py
```diff
@@ -9,18 +9,47 @@
...
+    result['kind'] = _kind(name)
+
+    # All assignments have the filename as a there toplevel parent.
+    if parent and parent.parent():
+        result['containerName'] = parent.name
```
gatesn, line 41: Ultimately this is the crucial line of this PR right?

pyls/plugins/symbols.py
```diff
@@ -9,18 +9,47 @@
...
+
+    return result
+
+
+def _range(name):
+    definition = name._name.tree_name.get_definition()
+    (start_line, start_column) = definition.start_pos
+    (end_line, end_column) = definition.end_pos
```
gatesn, line 49: What's the reasoning behind using a different method of accessing start/end positions? Does Jedi return different values?

gatesn (commented):
Thanks for making this PR. Keen to support the outline view, but think there the LS should respect the protocol as much as possible and push the filtering of symbols back on the client.

Do you mind if I take over this PR?

lgeiger:
🆒 Feel free to take over this PR. Sorry for my Javascript style Python code 😄 

This is mostly inspired by how [Nuclide does it's outline view](https://github.com/facebook/nuclide/blob/master/pkg/nuclide-python-rpc/python/outline.py).

pyls/plugins/symbols.py
lgeiger: Happy to have a config option for this.
Though I don't know any other language server that displays parameters and assignments inside a function. So this is mostly to bring `pyls` inline with other language servers.

Displaying all function parameters and assignments becomes messy very fast and in my opinion mitigates most of the advantage of having a outline view.

pyls/plugins/symbols.py
lgeiger: This in combination with the correct end positions of the definitions is the key here.

pyls/plugins/symbols.py
lgeiger: I'm also not happy using a private method here but Jedi only returns the start positions otherwise.
This results in problems with frontends not correctly displaying the structure. See https://github.com/atom/atom-languageclient/issues/116

pyls/plugins/symbols.py
gatesn: Other language servers behaviour is irrelevant in my opinion. The protocol is the source of truth here. 

However, since there is no SymbolKind for a function parameter, i agree it makes sense not to return them. Happy to filter those out. But underscore assignments and global assignments should (and could) be filtered by the client.

pyls/plugins/symbols.py
gatesn: Ah, is that the fix to the `_range` function?

lgeiger: Yes, this came mostly from the discussion at https://github.com/atom/atom-languageclient/issues/116

pyls/plugins/symbols.py
lgeiger: 👍 for not filtering out underscore assignments. It's just a thing I saw Nuclide do which felt neat.

> However, since there is no SymbolKind for a function parameter, i agree it makes sense not to return them.

I agree. That's how I got started doing all the filtering.
When only removing the function parameter from the outline view it feels strange when having a function like:
```python
def f(a):
    b = a
    return b
```
The outline view would show:
```bash
f
  a # only if also returning the parameter
  b
```
Which in my opinion is strange since `b` isn't a real property of `f`.

I think it is OK for the language server to make some language specific decisions what values to return.

Implementations in multi language editors like Atom or VSCode cannot easily make language agnostic decisions about displaying. So I think it's OK to provide some sensible defaults from the language server side.

gatesn:
Closed for #176

PR #176, opened 2017-10-28
Hierarchical symbols

gatesn:
@lgeiger want to decide on what the symbols look like:

```
import sys

a = 'hello'

class B:
    def __init__(self):
        x = 2
        self.y = x

def main(x):
    y = 2 * x
    return y
```

IMO:

- `sys`
- `a`
- `B`
- `B.__init__`
- `B.__init__.x`
- `main`
- `main.y`

### DIFF
--- a/pyls/plugins/symbols.py
+++ b/pyls/plugins/symbols.py
@@ -10,7 +10,6 @@ log = logging.getLogger(__name__)
 def pyls_document_symbols(config, document):
     all_scopes = config.plugin_settings('jedi_symbols').get('all_scopes', True)
     definitions = document.jedi_names(all_scopes=all_scopes)
-    print definitions[0].name, definitions[0].type
     return [{
         'name': d.name,
         'containerName': _container(d),
--- a/pyls/plugins/symbols.py
+++ b/pyls/plugins/symbols.py
@@ -22,11 +22,11 @@ def pyls_document_symbols(config, document):
 
 
 def _include_def(definition):
+    log.info("Definition %s has parent %s", definition.name, definition.parent().name)
     return (
         # Don't tend to include parameters as symbols
         definition.type != 'param' and
-        # Only include symbols that have a parent
-        _container(definition) is not None and
+        # Unused vars should also be skipped
         definition.name != '_' and
         _kind(definition) is not None
     )
@@ -36,9 +36,12 @@ def _container(definition):
     try:
         # Jedi sometimes fails here.
         parent = definition.parent()
-        return parent.name
+        # Here we check that a grand-parent exists to avoid declaring symbols
+        # as children of the module.
+        if parent.parent():
+            return parent.name
     except:
-        parent = None
+        pass
 
 
 def _range(definition):
--- a/pyls/plugins/symbols.py
+++ b/pyls/plugins/symbols.py
@@ -9,42 +9,40 @@ log = logging.getLogger(__name__)
 @hookimpl
 def pyls_document_symbols(config, document):
     all_scopes = config.plugin_settings('jedi_symbols').get('all_scopes', True)
-    names = document.jedi_names(all_scopes=all_scopes)
-    return list(filter(None, [serialize_name(n, document.uri) for n in names]))
+    definitions = document.jedi_names(all_scopes=all_scopes)
+    print definitions[0].name, definitions[0].type
+    return [{
+        'name': d.name,
+        'containerName': _container(d),
+        'location': {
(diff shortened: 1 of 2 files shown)


lgeiger:
Thanks for continuing the efforts!

This is how it looks in Atom:
bildschirmfoto 2017-10-28 um 20 53 35

The filename is `jedie.py`. [This line](https://github.com/palantir/python-language-server/pull/175/files#diff-cee448b8821a222944d7e6f2ca332ddaR40) used to prevent it from always showing.

Personally I'd prefer to not show `x` any `y` per default because things may become pretty cluttered on large files, but I'm happy to go with this and let more people chime in on the discussion.
I think it's OK to go with this for now and see if people complain about `pyls` showing too many symbols. Since `pyls` is pre 1.0 we can always iterate later.

`pyls/plugins/symbols.py` looks like this:
bildschirmfoto 2017-10-28 um 21 02 45

pyls/plugins/symbols.py
```diff
@@ -9,42 +9,40 @@
...
-         (parent and parent.parent()))) or name.type == 'param':
-        return None
-    else:
-        result['name'] = name.name
-
-    result['location'] = {'uri': uri, 'range': _range(name)}
-    result['kind'] = _kind(name)
-
-    # All assignments have the filename as a there toplevel parent.
-    if parent and parent.parent():
```
lgeiger: We still need this check, otherwise the filename is always the top most symbol.

gatesn:
I added the check to remove the top-level module name. But there's not sensible way to get rid of x and y without excluding things like a. So I'm going to leave it in for now.

gatesn:
@lgeiger I just had a thought... the names should be fully qualified. Otherwise a variable defined in constructors of two classes (__init__ method) can't be disambiguated.

lgeiger:
> I added the check to remove the top-level module name.

👍 

>  But there's not sensible way to get rid of x and y without excluding things like a. So I'm going to leave it in for now.

That's totally possible. That's what [this check did](https://github.com/palantir/python-language-server/pull/176/commits/55913e936e60b732aea05d301e9344d391eb42c8#diff-cee448b8821a222944d7e6f2ca332ddaR31).


> @lgeiger I just had a thought... the names should be fully qualified. Otherwise a variable defined in constructors of two classes (init method) can't be disambiguated.

I don't know if I fully understand you, but I think they can be distinguished easily by the indentation in the UI since the `__init__` methods hase different parents.

gatesn:
What I mean is that `a` will have containerName `__init__`, as will `b`.

gatesn:
I think we already make that check right? https://github.com/palantir/python-language-server/pull/176/files#diff-cee448b8821a222944d7e6f2ca332ddaR28

[edit] I see what's missing. Let me fix.

lgeiger:
> What I mean is that a will have containerName __init__, as will b.

Yes, that's why the correct start and end positions are important for the UI to figure out where they belong.

gatesn:
Got it, the positions are used in addition to containerName. 

Also, on second thoughts, a variable local to a function is still a symbol. So going to leave them.

lgeiger:
> I think we already make that check right?

No, the check [here](https://github.com/palantir/python-language-server/pull/176/commits/55913e936e60b732aea05d301e9344d391eb42c8#diff-cee448b8821a222944d7e6f2ca332ddaR31) also tests if the `statement` is defined inside a function and omits it. If `parent.parent()` isn't defined it's in the global scope, otherwise it was defined inside a function.

> Got it, the positions are used in addition to containerName.

Yeah, it's not exactly the most elegant way to infer hierarchy 😄 

> Also, on second thoughts, a variable local to a function is still a symbol. So going to leave them.

I'm OK with this. I don't know what the canonical way would be.

gatesn:
Ok, sounds sounds like we're in agreement. I also don't mind if the client excludes variables whose container symbol is a function. Most outline views tend to be expandable trees though, so I wonder if it can just be solved with default level of expanded subtrees.

Anyway, going to merge to get this out the door. Thanks for all your help / opinions!

lgeiger:
> so I wonder if it can just be solved with default level of expanded subtrees.

It's not so easy to solve this with default levels since one would also exclude class methods then.

> Anyway, going to merge to get this out the door. Thanks for all your help / opinions!

Cool looking forward to the next release.

bhack:
Do you think that we will have a new tagged release for this soon?

bhack:
@lgeiger The is a new release now.

PR #198, opened 2017-12-06
Fix line length issue with jedi

### LINKED ISSUE #193: Completion requests with column positions greater than the line length don't work (LSP implementation issue)
The LSP protocol document says that `character` values for the position, used for methods like `textDocument/completion` should accept values greater than the length of a line. https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#position

```
* If the character value is greater than the line length it defaults back to the
* line length.
```

However, validation in `jedi` causes an exception to be thrown.

```
558.368137 on 1: Dropping message '2017-11-21 17:22:37,470 UTC - ERROR - jsonrpc.manager - API Exception: {'type': 'ValueError', 'args': ('`column` parameter is not in a valid range.',), 'message': '`column` parameter is not in a valid range.'}
Traceback (most recent call last):
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/jsonrpc/manager.py", line 112, in _get_responses
    result = method(*request.args, **request.kwargs)
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/pyls/language_server.py", line 66, in wrapped
    return func(*args, **kwargs)
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/pyls/python_ls.py", line 138, in m_text_document__completion
    return self.completions(textDocument['uri'], position)
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/pyls/python_ls.py", line 68, in completions
    position=position
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/pyls/_utils.py", line 52, in race_hooks
    for impl, result in pool.imap_unordered(_apply, impls, chunksize=1):
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 735, in next
    raise value
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/pyls/_utils.py", line 47, in _apply
    return impl, impl.function(**kwargs)
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/pyls/plugins/jedi_completion.py", line 12, in pyls_completions
    definitions = document.jedi_script(position).completions()
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/pyls/workspace.py", line 265, in jedi_script
    return jedi.Script(**kwargs)
  File "/home/w0rp/Downloads/python-ls-project/ve/lib/python3.6/site-packages/jedi/api/__init__.py", line 102, in __init__
    raise ValueError('`column` parameter is not in a valid range.')
ValueError: `column` parameter is not in a valid range.
```

The ideal way to fix this would be to remove the validation in `jedi`, and make it use `column = min(column, line_len)` instead.

gatesn:
Fixes #193

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -263,6 +263,6 @@ class Document(object):
             'sys_path': self._sys_path
         }
         if position:
-            kwargs['line'] = position['line'] + 1
-            kwargs['column'] = position['character']
+            kwargs['line'] = min(position['line'] + 1, len(self.lines)) or 1
+            kwargs['column'] = min(position['character'], len(self.lines[kwargs['line'] - 1]) - 1)
         return jedi.Script(**kwargs)
--- a/test/plugins/test_completion.py
+++ b/test/plugins/test_completion.py
@@ -41,6 +41,13 @@ def test_jedi_completion():
     assert items[0]['label'] == 'isabs(s)'
 
 
+def test_jedi_completion_bad_position():
+    bad_position = {'line': 100, 'character': 1500}
+    doc = Document(DOC_URI, DOC)
+    # Does not raise
+    pyls_jedi_completions(doc, bad_position)
+
+
 def test_rope_completion():
     # Over 'i' in os.path.isabs(...)
     com_position = {'line': 1, 'character': 15}


gatesn:
Not sure this is a great solution given that a bad position will still cause exceptions for other methods on Document

Issue #199, opened 2017-12-08
Support non-file URIs


dirathea:
# Summary
I've tried to use  PYLS on Monaco Editor using [Monaco Language Client](https://github.com/TypeFox/monaco-languageclient) with the following configuration

```
this.editor.setModel(monaco.editor.createModel(exampleCode, 'python', monaco.Uri.parse('inmemory://dummy.py')););
    const services = createMonacoServices(this.editor);
```
And everything that [this blog](http://typefox.io/teaching-the-language-server-protocol-to-microsofts-monaco-editor) explained.

Everything works as expected but the auto complete. When I do `ctrl + space` on the editor, there's only "No Suggestions."  and I saw the PYLS logs error as follows

```
PYLS Server: 2017-12-08 09:56:21,029 UTC - ERROR - jsonrpc.manager - API Exception: {'message': "'Folder' object has no attribute 'read'", 'args': ("'Folder' object has no attribute 'read'",), 'type': 'AttributeError'}
Traceback (most recent call last):
  File "myVirtualEnv/lib/python2.7/site-packages/jsonrpc/manager.py", line 112, in _get_responses
    result = method(*request.args, **request.kwargs)
  File "myVirtualEnv/lib/python2.7/site-packages/pyls/language_server.py", line 64, in wrapped
    return func(*args, **kwargs)
  File "myVirtualEnv/lib/python2.7/site-packages/pyls/python_ls.py", line 140, in m_text_document__completion
    return self.completions(textDocument['uri'], position)
  File "myVirtualEnv/lib/python2.7/site-packages/pyls/python_ls.py", line 70, in completions
    position=position
  File "myVirtualEnv/lib/python2.7/site-packages/pyls/_utils.py", line 113, in race_hooks
    for impl, result in pool.imap_unordered(_apply, impls, chunksize=1):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/multiprocessing/pool.py", line 668, in next
    raise value
AttributeError: 'Folder' object has no attribute 'read'
```

Could you help me? is it about the monaco URI that I pass ? `inmemory://dummy.py`

gatesn:
I think there is an issue somewhere about supporting URIs that have schemes other than “file”. I suspect this is what you’re hitting. I believe the SourceGraph language server has better support for things like this right now. 

https://github.com/sourcegraph/python-langserver

exiaohu:
[...]

PR #200, opened 2017-12-09
[Auto Complete] force return None if empty array is returned

dirathea:
# Summary
`race_hooks` says it's `returning the first non-None result`, but in the implementations on auto complete the `rope_completion` return empty array instead of None, while `jedi_completion` might return non empty value.
This PR is to force return `None` for empty result from `rope_completion` to let `jedi_completion` finished, and might gives better completions.

related : #199

### DIFF
--- a/pyls/plugins/rope_completion.py
+++ b/pyls/plugins/rope_completion.py
@@ -41,9 +41,7 @@ def pyls_completions(document, position):
             'sortText': _sort_text(d)})
     definitions = new_definitions
     log.debug('Rope finished')
-    if len(definitions) == 0:
-        return None
-    return definitions
+    return definitions or None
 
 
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @dirathea! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn:
Thanks for spotting this and filing a PR! Would you be able to change the condition to `if not definitions` in keeping with the repository style? 

Other than that, looks great.

gatesn:
Actually, I think `return definitions or None` is a little cleaner.

PR #211, opened 2017-12-26
Fix pycodestyle iterable options

### LINKED ISSUE #239: Fix ignored and select settings interface with pycodestyle
On https://github.com/PyCQA/pycodestyle/pull/722 they refused to fix their interface. When passing the list arguments as `ignore` and `select` settings to `pycodestyle`, it is required to pass a python list as `["E201", "E501"]`, instead of a string `"E201,E501"`, otherwise they will cause the issue pointed on: https://github.com/tomv564/LSP/issues/244#issuecomment-358753274

lgeiger:
I'm trying to support the new pycodestyle config options in the `ide-python` Atom package: https://github.com/lgeiger/ide-python/pull/58

Unfortunately with `pyls` 0.12.1 ignoring multiple error codes isn't working for me.

~~This PR adds a failing test case illustrating the behaviour. Any help with this issue would be really appreciated.~~

This fixes #239 by reverting #220.

Happy holidays.

### DIFF
--- a/pyls/plugins/pycodestyle_lint.py
+++ b/pyls/plugins/pycodestyle_lint.py
@@ -12,12 +12,12 @@ def pyls_lint(config, document):
     log.debug("Got pycodestyle settings: %s", settings)
 
     opts = {
-        'exclude': ','.join(settings.get('exclude') or []),
-        'filename': ','.join(settings.get('filename') or []),
+        'exclude': settings.get('exclude'),
+        'filename': settings.get('filename'),
         'hang_closing': settings.get('hangClosing'),
-        'ignore': ','.join(settings.get('ignore') or []),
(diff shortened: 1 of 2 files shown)


evandrocoan:
@lgeiger, This should have been fixed by https://github.com/palantir/python-language-server/pull/234

lgeiger:
> This should have been fixed by #234

👍I rebased. Let's see what CI thinks.

lgeiger:
Thanks @evandrocoan for pointing this out.

I had to revert #220 to fix #239. Now the tests should pass.

lgeiger:
We've had quite a bit of bugfixes lately. @gatesn Is there anything holding back the next release?

PR #212, opened 2017-12-28
Completion: Remove pluggy racing and disable Rope by default

### LINKED ISSUE #189: Split out rope a bit more until it's stable
I've seen quite a few issues from the rope implementation. Probably going to revert / disable it by default.

Will keep pre-loading some known large modules, perhaps in another thread, in order to warm up Jedi. Also going to remove pluggy racing, since Jedi's cache isn't thread safe.

lgeiger:
This allows the user to choose between Rope and Jedi for completions.

It also remove the pluggy racing since Jedi's cache isn't thread safe.

Closes #189

### DIFF
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -23,12 +23,7 @@ def pyls_commands(config, workspace):
 
 
 @hookspec
-def pyls_jedi_completions(config, workspace, document, position):
-    pass
-
-
-@hookspec
-def pyls_rope_completions(config, workspace, document, position):
+def pyls_completions(config, workspace, document, position):
     pass
 
 
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -23,7 +23,12 @@ def pyls_commands(config, workspace):
 
 
 @hookspec
-def pyls_completions(config, workspace, document, position):
+def pyls_jedi_completions(config, workspace, document, position):
+    pass
+
+
+@hookspec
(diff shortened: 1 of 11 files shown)


pyls/hookspecs.py
```diff
@@ -23,7 +23,12 @@ def pyls_commands(config, workspace):
 
 
 @hookspec
-def pyls_completions(config, workspace, document, position):
+def pyls_jedi_completions(config, workspace, document, position):
```
gatesn: Why have two plugin hooks? Why not just control which implementation runs using the configuration?

gatesn (commented):
i think we should leave just the single completions hook, and users can enable one, the other, or both.

pyls/hookspecs.py
lgeiger: I don't think users should be able to use both completion plugins side by side since this will probably result in duplicated completion results and other weird behaviour, or am I missing something.

I can refactor it to only use on completion hook which runs either jedi or rope depending on the config. What do you think?

gatesn:
Apologies for the slow review here, I’m still on vacation until the 9th January. Hope you understand!

lgeiger:
> Apologies for the slow review here, I’m still on vacation until the 9th January. Hope you understand!

No problem. Happy holidays!

gatesn:
@lgeiger do you mind if I take over this PR?

lgeiger:
>  do you mind if I take over this PR?

Go ahead!

PR #216, opened 2018-01-04
Fix character normalization

lgeiger:
The language server will throw a exception if a request is sent from an empty document because `self.lines == []` and therefor `self.lines[0]` will throw `IndexError: list index out of range`.

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -267,6 +267,6 @@ class Document(object):
 
             # Normalise the position as per the LSP that accepts character positions > line length
             # https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#position
-            line_len = len(self.lines[position['line']])
-            kwargs['column'] = min(position['character'], line_len - 1)
+            max_column = len(self.lines[position['line']]) - 1 if len(self.lines) > position['line'] else 0
+            kwargs['column'] = min(position['character'], max_column)
         return jedi.Script(**kwargs)
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -264,9 +264,5 @@ class Document(object):
         }
         if position:
             kwargs['line'] = position['line'] + 1
-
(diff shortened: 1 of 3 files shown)


pyls/workspace.py
```diff
@@ -267,6 +267,6 @@ def jedi_script(self, position=None):
 
             # Normalise the position as per the LSP that accepts character positions > line length
             # https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#position
-            line_len = len(self.lines[position['line']])
-            kwargs['column'] = min(position['character'], line_len - 1)
+            max_column = len(self.lines[position['line']]) - 1 if len(self.lines) > position['line'] else 0
```
gatesn: Doesn't this also end up with passing 0 into Jedi? Is there a test you could add perhaps?

lgeiger: > Doesn't this also end up with passing 0 into Jedi?

It does, but does Jedi have a problem when 0 is passed to it?

This fix only prevents `IndexError: list index out of range` on line 270 because `self.lines` may be an empty list.

Sorry for not adding tests, I'll do that.

lgeiger: OK, I added a small test.

test/test_utils.py
```diff
@@ -49,3 +49,9 @@ def test_merge_dicts():
         {'a': True, 'b': {'x': 123, 'y': {'hello': 'world'}}},
         {'a': False, 'b': {'y': [], 'z': 987}}
     ) == {'a': False, 'b': {'x': 123, 'y': [], 'z': 987}}
+
+
+def test_clip_column():
+    assert _utils.clip_column(5, ['123'], 0) == 2
+    assert _utils.clip_column(2, ['\n', '123'], 1) == 2
+    assert _utils.clip_column(0, [], 0) == 0
```
lgeiger, line 57: This test would previously throw, yet this state actually occurs when inside a empty document.

gatesn (approved):

PR #222, opened 2018-01-10
Remove pluggy racing

### LINKED ISSUE #189: Split out rope a bit more until it's stable

gatesn:
@lgeiger does this work for you?

Fixes #212 
Fixes #189


lgeiger:
With this the user is allowed to enable both completion providers and would receive duplicated completions.

I think [this](https://github.com/palantir/python-language-server/pull/212/files#diff-942a8b6d9b93cba7f0d40bd9db4a4d2fR36) would be a nicer configuration interface for users. What do you think.

I can test this branch tomorrow and provide detailed feedback.

gatesn:
So we could either pass `@hookspec(firstresult=True)` to the completions hook, meaning we'll use the first non-None result. Or we merge and try to filter duplicates.

I'd rather not have config to switch between them because it means the user needs to understand the pros/cons of each. Also, it's similar behaviour to firstresult=True, but less powerful since there's no fallback.

The disadvantage of first result (and your solution too), is that it becomes impossible for a plugin to add extra completions into the mix. Which was the point of plugins in the first place.

lgeiger:
I think this is the most flexible solution for now. We can think about merging to filter duplicates in the future once people start using this.

As for the configuration, editor integrations can always decide for them selves to present a different config schema to the users.

lgeiger (approved):

PR #232, opened 2018-01-18
Created a string representation for the Config and ConfigSource

evandrocoan:
classes on pyls/config/config.py and source.py files.

Such representation is very helpful for debugging the application
as we can easily see the loaded configurations while running the
code.

### DIFF
--- a/pyls/config/config.py
+++ b/pyls/config/config.py
@@ -51,7 +51,7 @@ class Config(object):
             "_disabled_plugins: %s" % str(self._disabled_plugins),
         ]
         representation.extend(["_config_sources(%s): %s" % (item, self._config_sources[item])
-                for item in self._config_sources])
+                              for item in self._config_sources])
         return ", ".join(representation)
 
     @property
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @evandrocoan! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn:
I think this might pollute the logs a bit. Is there somewhere we could explicitly log the config, or changes to it, such that log lines are kept smaller?

evandrocoan:
> Is there somewhere we could explicitly log the config

I do not know.  Before using this to print the settings object, I was completely lost about what does yours settings objects where holding. This is not supposed to be print on the log on daily basis, but only for new project contributors which still learning how does this system is structured and its work flow.

There are already other lines much longer than these as the results from pycodestyle. So when these lines are coming I just disable line wrapping. However, these lines are not being printed anywhere on the log currently. To use this you need to put some log statement somewhere. The point of __str__ is to see the object attributes when you do not what they are currently holding.

gatesn:
Closing this out, I'm not sure we want to do this

PR #240, opened 2018-01-25
Fixed pydocstyle_lint.py showing one diagnostic error per line,

evandrocoan:
instead of show all of the at once.


gatesn (approved):

gatesn:
@evandrocoan `log.debug` do you think?

evandrocoan:
Makes more sense because if someone will be watching these values, is because if debugging something, instead of just wanting to be "informed" of something.

gatesn (approved):

gatesn:
@evandrocoan wanted to thank you for all your valuable contributions! Apologies if our code reviews aren't always as prompt as necessary.

PR #243, opened 2018-01-25
Replace the logging.basicConfig by a proper configuration

evandrocoan:
basicConfig: Do basic configuration for the logging system.
This function does nothing if the root logger already has handlers
configured. It is a convenience method intended for use by simple
scripts to do one-shot configuration of the logging package.

### DIFF
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -65,6 +65,7 @@ def _binary_stdio():
     PY3K = sys.version_info >= (3, 0)
 
     if PY3K:
+        # pylint: disable=no-member
         stdin, stdout = sys.stdin.buffer, sys.stdout.buffer
     else:
         # Python 2 on Windows opens sys.stdin in text mode, and
@@ -89,7 +90,7 @@ def _configure_logger(args):
             logging.config.dictConfig(json.load(f))
     else:
         log_file = args.log_file
-        formatter = logging.Formatter(LOG_FORMAT, style="%")
+        formatter = logging.Formatter(LOG_FORMAT)
         if log_file:
             log_handler = logging.handlers.RotatingFileHandler(
                 log_file, mode='a', maxBytes=50*1024*1024,
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -4,6 +4,7 @@ import json
 import logging
 import logging.config
 import sys
+
 from . import language_server
 from .python_ls import PythonLanguageServer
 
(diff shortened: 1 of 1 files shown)


pyls/__main__.py
```diff
@@ -95,3 +80,37 @@ def _binary_stdio():
...
+    root_logger = logging.root
+
+    if log_config:
+        with open(log_config, 'r') as f:
+            logging.config.dictConfig(json.load(f))
+    else:
+        log_file = args.log_file
+        formatter = logging.Formatter(LOG_FORMAT, style="%")
+        if log_file:
+            log_handler = logging.handlers.RotatingFileHandler(log_file, mode='a',
```
gatesn: Can you stick to repo parenthesis style? 
```
logging.handlers.RotatingFileHandler(
    log_file, mode='a', maxBytes=50*1024*1024, 
    backupCount=10, encoding=None, delay=0
)
```

pyls/__main__.py
```diff
@@ -95,3 +80,37 @@ def _binary_stdio():
...
+            log_handler = logging.StreamHandler()
+        log_handler.setFormatter(formatter)
+        root_logger.addHandler(log_handler)
+
+    # CRITICAL = 50
+    # FATAL = CRITICAL
+    # ERROR = 40
+    # WARNING = 30
+    # INFO = 20
+    # DEBUG = 10
```
gatesn: Why are these comments here? They don't seem to be relevant to the code

pyls/__main__.py
```diff
@@ -95,3 +80,37 @@ def _binary_stdio():
         stdin, stdout = sys.stdin, sys.stdout
 
     return stdin, stdout
+
+def configure_logger(args):
```
gatesn: Can you prefix function with underscore

evandrocoan: Sure.

pyls/__main__.py
evandrocoan: Oops.

pyls/__main__.py
evandrocoan: Informative about the debug levels. I put it there because I was confused about them when I saw you can only pass 3 levels, but I know the `logging` module has 6. So, looking at these values I can clearly see that we you set the WARNING level, you can also see the ERROR and other above it.

gatesn:
Some linting errors too:

```
************* Module pyls.__main__
C: 96, 0: Wrong continued indentation (add 43 spaces).
                    maxBytes=50*1024*1024, backupCount=10, encoding=None, delay=0)
                    ^                                          | (bad-continuation)
C:116, 0: Trailing newlines (trailing-newlines)
E: 93,20: Unexpected keyword argument 'style' in constructor call (unexpected-keyword-arg)
I: 75, 0: Useless suppression of 'no-member' (useless-suppression)
I: 75, 0: Useless suppression of 'import-error' (useless-suppression)
```

evandrocoan:
I should have run `tox` before. I am running it now.

pyls/__main__.py
gatesn: I think that's part of the logging API and doesn't need re-documenting here. Most logging frameworks would log at the given level or "above"

evandrocoan:
@gatesn, this linter is insane:

1. First it complained about missing 2 trailing new lines. 
1. Now it complains about a existent trailing new line.

PR #246, opened 2018-01-25
fix completion sorting

ferozco:
solves issue where hidden members `_*` appear before non hidden members

### DIFF
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -20,26 +20,25 @@ def pyls_completions(document, position):
 
 def _label(definition):
     if definition.type in ('function', 'method'):
-        params = ", ".join(param.name for param in definition.params)
-        return "{}({})".format(definition.name, params)
+        params = ', '.join(param.name for param in definition.params)
+        return '{}({})'.format(definition.name, params)
 
     return definition.name
 
 
 def _detail(definition):
-    return "builtin" if definition.in_builtin_module() else definition.parent().full_name or ""
+    return 'builtin' if definition.name.startswith('_') else definition.parent().full_name or ''
 
 
 def _sort_text(definition):
     """ Ensure builtins appear at the bottom.
     Description is of format <type>: <module>.<item>
     """
-    if definition.in_builtin_module():
-        # It's a builtin, put it last
+    if definition.name.startswith('_'):
+        # It's a 'hidden' func, put it next last
         return 'z' + definition.name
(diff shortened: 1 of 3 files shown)


pyls/plugins/jedi_completion.py
```diff
@@ -20,26 +20,25 @@ def pyls_completions(document, position):
...
-        return "{}({})".format(definition.name, params)
+        params = ', '.join(param.name for param in definition.params)
+        return '{}({})'.format(definition.name, params)
 
     return definition.name
 
 
 def _detail(definition):
-    return "builtin" if definition.in_builtin_module() else definition.parent().full_name or ""
+    return 'builtin' if definition.name.startswith('_') else definition.parent().full_name or ''
```
gatesn: I'd remove this entirely, not sure it's valid to say "__dict__" is a builtin, instead of being an attribute on the module.

`return definition.parent().full_name or ''`

pyls/plugins/jedi_completion.py
```diff
@@ -20,26 +20,25 @@ def pyls_completions(document, position):
...
     """
-    if definition.in_builtin_module():
-        # It's a builtin, put it last
+    if definition.name.startswith('_'):
+        # It's a 'hidden' func, put it next last
         return 'z' + definition.name
 
-    if definition.name.startswith("_"):
-        # It's a 'hidden' func, put it next last
+    if definition.in_builtin_module():
```
gatesn: I don't trust the `definition.in_builtin_module` function at all. It checks if the module is from a compiled file, which doesn't necessarily mean its a builtin. It might just not come with sources. Remove this whole block I think

pyls/plugins/rope_completion.py
```diff
@@ -52,11 +52,10 @@ def _sort_text(definition):
     """ Ensure builtins appear at the bottom.
     Description is of format <type>: <module>.<item>
     """
-    if definition.scope == 'builtin':
-        return 'z' + definition.name
-
     if definition.name.startswith("_"):
         # It's a 'hidden' func, put it next last
+        return 'z' + definition.name
+    elif definition.scope == 'builtin':
```
gatesn, line 58: Can't really comment on this as don't know much about Rope, trust it's sensible though.

pyls/plugins/jedi_completion.py
ferozco: isn't dict a built in?

PR #247, opened 2018-01-25
Fix pydocstyle reading from sys.argv

### LINKED ISSUE #228: pydocstyle linter doesn't ignore `sys.argv`
When running in TCP mode, pydocstyle complains that it doesn't understand the `--tcp` argument. Presumably it's reading directly from `sys.argv`

gatesn:
Fixes #228 

c.f. https://github.com/PyCQA/pydocstyle/issues/299

### DIFF
--- a/pyls/plugins/pydocstyle_lint.py
+++ b/pyls/plugins/pydocstyle_lint.py
@@ -1,5 +1,8 @@
 # Copyright 2017 Palantir Technologies, Inc.
+import contextlib
 import logging
+import sys
+
 import pydocstyle
 from pyls import hookimpl, lsp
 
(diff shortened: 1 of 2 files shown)


pyls/plugins/pydocstyle_lint.py
```diff
@@ -19,8 +22,10 @@ def pyls_settings():
 @hookimpl
 def pyls_lint(document):
     conf = pydocstyle.config.ConfigurationParser()
-    conf.parse()
-    conf._arguments = [document.path]
+
+    with _patch_sys_argv([document.path]):
+        # TODO(gatesn): We can add more pydocstyle args here from our pyls config
+        conf.parse()
```
evandrocoan, line 28: We can also add the options `ignore` and `select` which are ignored when set on the user configuration file: 
```javascript
"pyls": {
    "plugins": {
        "pydocstyle": {
            "enabled": true,
            "ignore": [
                "D202",
                "D208"
            ]
        }
    }
}
```

gatesn: follow-up in #249

Issue #230, opened 2018-01-17
Maybe we should prevent creating a .ropeproject/config.py


segevfiner:
When you use pyls today, it ends up creating a `.ropeproject/config.py` in any project that you open with it. This might make sense for users who use rope directly, but less so for someone unfamiliar with it, who just tried to use pyls and ends up with a new untracked file in his VCS.

Maybe something should be done to prevent it's creation by pyls? (If at all possible...)

exploide:
I also thought of this and according to the documentation it is possible to suppress folder creation: https://github.com/python-rope/rope/blob/master/docs/library.rst#ropeproject-folder

However, I'm not familiar with rope and do not know what exactly is done with this folder and if it is reasonable to disable it or if it would lead to performance problems or whatever.

segevfiner:
I don't know either. Hence the "maybe" in the title. That's what the
documentation has to say https://github.com/python-rope/rope/blob/master/docs/overview.rst#ropeproject-folder.

It seems it holds configuration, history and caches.

It can always be made configurable rather than outright disabling it...

בתאריך 18 בינו' 2018 16:21,‏ "Jannik Vieten" <notifications@github.com> כתב:

evandrocoan:
This configuration file is extremely annoying. I would never use this tool until this madness is removed. Next thing to do is to find and kill this trash file creation. 

If I would like this damn thing on my project, I can create the damn file myself from a template.

segevfiner:
I guess the first two things that can be done are making the Rope `Project` initialization lazy rather than at the initialization of pyls (I think it's in workspace.py), and adding a configuration value for the location of `.ropeproject` that can also include some replacement variables like `$PROJECT` or `{PROJECT}` etc that can be used to relocate it to a project dependent location outside of the project directory.

I wonder what other projects using Rope do...

evandrocoan:
I am attempting to create a setting on https://github.com/palantir/python-language-server/pull/253 to disable the folder creation. You guys can test and see if it is working correctly for you.

mcepl:
If you disable creating this folder, please, make it optional. There are some fun things I do with ``~/.ropeproject/config.py``, which I would miss.

gatesn:
I believe this PR means you won't get a rope project if it's not installed: https://github.com/palantir/python-language-server/pull/301

PR #253, opened 2018-01-29
Created the rope.create_folder setting to not create the

evandrocoan:
.ropeproject folder and config.py file. Related to:
1. https://github.com/python-rope/rope/issues/234 you can even tell rope not to create .ropeproject folder
1. https://github.com/palantir/python-language-server/issues/230 Maybe we should prevent creating a .ropeproject/config.py

The setting can be created just like the others:
```json
			"settings":
			{
				"pyls":
				{
					"plugins":
					{
						"pycodestyle":
						{
							"enabled": false
						},
						"rope":
						{
							"create_folder": false
						},
					}
				}
			},
```

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -9,7 +9,7 @@ import pkgutil
 
 import jedi
 from rope.base import libutils
-from rope.base.project import Project
+from rope.base.project import Project, get_no_project
 
 from . import lsp, uris, _utils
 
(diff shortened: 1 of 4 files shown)


pyls/workspace.py
```diff
@@ -74,21 +74,24 @@ class Workspace(object):
...
+    def __init__(self, root_uri, lang_server):
         self._root_uri = root_uri
         self._root_uri_scheme = uris.urlparse(self._root_uri)[0]
         self._root_path = uris.to_fs_path(self._root_uri)
         self._docs = {}
         self._lang_server = lang_server
 
         # Whilst incubating, keep private
-        self.__rope = Project(self._root_path)
+        if self._lang_server.config.plugin_settings('rope').get('create_folder', False):
```
evandrocoan, line 85: Get the user setting here does not work because the user settings are only loaded by the server after this function to be called.

PR #254, opened 2018-01-30
Replaced RotatingFileHandler by ConcurrentRotatingFileHandler

evandrocoan:
Replaced RotatingFileHandler by ConcurrentRotatingFileHandler
module concurrent-log-handler because when there are multiple pyls
server's running the file rotation fails.

### DIFF
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -5,6 +5,8 @@ import logging
 import logging.config
 import sys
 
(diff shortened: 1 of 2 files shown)


gatesn:
I don't think we want to have multiple language servers logging to the same file. Whoever's running pyls should construct a sensible log path. Or we should support putting the pid in the log filename

PR #260, opened 2018-02-14
Add support for asynchronously handling requests

ferozco:
Previously we had a single thread that would handle and subsequently respond to each message in order. This presented performance problems since long running handlers would cause the server to hang from the perspective of the client. Furthermore,  prevented a handler from sending (and receiving) multiple message.

This change allows handlers to be executed either synchronously or asynchronously with the choice being left to the specific hookimpl.


pyls/message_manager.py
```diff
@@ -0,0 +1,116 @@
...
+from jsonrpc.jsonrpc import JSONRPCRequest
+from jsonrpc.exceptions import (
+    JSONRPCInvalidRequestException,
+)
+
+
+log = logging.getLogger(__name__)
+
+
+class MessageManager(object):
```
gatesn: JSONRPCServer ?

pyls/message_manager.py
```diff
@@ -0,0 +1,116 @@
...
+        self.write_lock = threading.Lock()
+
+    def close(self):
+        with self.write_lock:
+            self.wfile.close()
+            self.rfile.close()
+
+    def get_messages(self):
+        """
+        Generator that produces well structured JSON RPC message.
```
gatesn: Docs start on same line as `"""` without a space. So  `"""Generator that ...`

pyls/message_manager.py
```diff
@@ -0,0 +1,116 @@
...
+            message: received message
+
+        Note:
+            This method is not thread safe and should only invoked from a single thread
+        """
+        while not self.rfile.closed:
+            request_str = self._read_message()
+
+            if request_str is None:
+                # log.error("failed to read message")
```
gatesn: Should we break here? Looks like we might be at EOF

pyls/message_manager.py
```diff
@@ -0,0 +1,116 @@
...
+                continue
+            if isinstance(request_str, bytes):
+                request_str = request_str.decode("utf-8")
+
+            try:
+                try:
+                    message_blob = json.loads(request_str)
+                    message = JSONRPCRequest.from_data(message_blob)
```
gatesn: Does this support batch requests?

pyls/message_manager.py
```diff
@@ -0,0 +1,116 @@
...
+                except JSONRPCInvalidRequestException:
+                    # work around where JSONRPC20Reponse expects _id key
+                    message_blob['_id'] = message_blob['id']
```
gatesn: Why are we doing this in an except block? Surely if `'id' in message_blob` then we should always just set `_id`?

pyls/message_manager.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def write_message(self, message):
+        """ Write message to out file descriptor
+
+        Args:
+            message (dict): body of the message to send
+        """
+        with self.write_lock:
+            if self.wfile.closed:
+                return
```
gatesn: Shouldn't we throw here? The client would see this as a successful write. There's also a race if the wfile is closed after this line but before the wfile.write call.

Just remove this check

pyls/python_ls.py
```diff
@@ -1,44 +1,105 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import logging
-from . import lsp, _utils
+import re
+import socketserver
+
+from . import lsp, _utils, uris
```
gatesn: Order alphabetically

pyls/python_ls.py
```diff
@@ -1,44 +1,105 @@
...
+        super(_StreamHandlerWrapper, self).setup()
+        # pylint: disable=no-member
+        self.delegate = self.DELEGATE_CLASS(self.rfile, self.wfile)
+
+    def handle(self):
+        self.delegate.handle()
+
+
+def start_tcp_lang_server(bind_addr, port, handler_class):
+    if not issubclass(handler_class, ):
```
gatesn: This seems... weird

pyls/python_ls.py
```diff
@@ -1,44 +1,105 @@
...
+        log.info('Serving %s on (%s, %s)', handler_class.__name__, bind_addr, port)
+        server.serve_forever()
+    finally:
+        log.info('Shutting down')
+        server.server_close()
+
+
+def start_io_lang_server(rfile, wfile, handler_class):
+    if not issubclass(handler_class, PythonLanguageServer):
+        raise ValueError('Handler class must be a subclass of JSONRPCServer')
```
gatesn: Error message doesn't match is subclass check

pyls/python_ls.py
```diff
@@ -1,44 +1,105 @@
...
+        self.rpc_manager = JSONRPCManager(MessageManager(rx, tx), self.handle_request)
+        self.workspace = None
+        self.config = None
+        self._dispatchers = []
 
-    # Set of method dispatchers to query
-    _dispatchers = []
+    def start(self):
+        """Entry point for the server"""
+        self.rpc_manager.consume_requests()
```
gatesn: Think it makes more sense to pass `handle_request` here, rather than in constructor or JSONRPCManager

pyls/python_ls.py
```diff
@@ -1,44 +1,105 @@
...
 
-    def __getitem__(self, item):
-        """Override the method dispatcher to farm out any unknown messages to our plugins."""
-        try:
-            return super(PythonLanguageServer, self).__getitem__(item)
-        except KeyError:
-            log.debug("Checking dispatchers for %s: %s", item, self._dispatchers)
+    def handle_request(self, method, params):
+        """Provides calllables to handle message requests
```
gatesn: Typo

pyls/python_ls.py
```diff
@@ -1,44 +1,105 @@
...
+
+        Args:
+            method (str):
```
gatesn: Remove docs if they're not filled out

pyls/python_ls.py
```diff
@@ -1,44 +1,105 @@
...
-                    return dispatcher.__getitem__(item)
-                except KeyError:
-                    pass
-        raise KeyError("Unknown item %s" % item)
+                if method_call in dispatcher:
+                    return dispatcher[method_call](**params)
 
-    def _hook_caller(self, hook_name):
-        return self.config.plugin_manager.subset_hook_caller(hook_name, self.config.disabled_plugins)
+        raise KeyError
```
gatesn: Can you pass a more descriptive message here.

pyls/python_ls.py
```diff
@@ -66,15 +127,40 @@ def capabilities(self):
...
+    def m_initialize(self, **kwargs):
+        log.debug('Language server initialized with %s', kwargs)
+        if 'rootUri' in kwargs:
+            root_uri = kwargs['rootUri']
+        elif 'rootPath' in kwargs:
+            root_path = kwargs['rootPath']
+            root_uri = uris.from_fs_path(root_path)
+        else:
+            root_uri = ''
+        init_opts = kwargs.get('initializationOptions')
```
gatesn: Can you `get('init...', {})`

pyls/python_ls.py
```diff
@@ -66,15 +127,40 @@ def capabilities(self):
...
+
+        self.workspace = Workspace(root_uri, rpc_manager=self.rpc_manager)
         self.config = config.Config(root_uri, init_opts)
         self._dispatchers = self._hook('pyls_dispatchers')
         self._hook('pyls_initialize')
 
+        # Get our capabilities
+        return {'capabilities': self.capabilities()}
+
+    def m___cancel_request(self, **kwargs):
```
gatesn: `(self, id=None, **_kwargs):`

pyls/python_ls.py
```diff
@@ -124,6 +210,9 @@ def rename(self, doc_uri, position, new_name):
     def signature_help(self, doc_uri, position):
         return self._hook('pyls_signature_help', doc_uri, position=position)
 
+    def m__cancel_request(self, **kwargs):
```
gatesn: Which cancel request is actually used? Why are there 2 underscores in this and 3 in the other?

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+log = logging.getLogger(__name__)
+
+RESPONSE_CLASS_MAP = {
+    "1.0": JSONRPC10Response,
+    "2.0": JSONRPC20Response
+}
+
+
+class JSONRPCManager(object):
+    """ Implementation of the Microsoft VSCode Language Server Protocol
```
gatesn: Not sure this doc string is true

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+
+log = logging.getLogger(__name__)
+
+RESPONSE_CLASS_MAP = {
+    "1.0": JSONRPC10Response,
+    "2.0": JSONRPC20Response
+}
+
+
+class JSONRPCManager(object):
```
gatesn, line 19: Something needs to be called Server!! 😄

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+
+         Args:
+             method (str): The method name of the notification to send
+             params (dict): The payload of the notification
+         """
+        log.debug('Notify %s %s', method, params)
+        notification = JSONRPC20Request(method=method, params=params)
+        self._message_manager.write_message(notification.data)
+
+    def cancel(self, request_id):
```
gatesn, line 71: You don't want/need this. Instead, you should add a done callback in the `call` method that checks if the future is cancelled, and if so sends the cancel message. That way a user can call `manager.call('method').cancel()`

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+        """Execute corresponding handler for the recieved request
+
+        Args:
+            request (JSONRPCBaseRequest): request to act upon
+
+        Note:
+            requests are handled asynchronously if the handler returns a callable, otherwise they are handle
+            synchronously by the main thread
+        """
+        if self._shutdown and request.method != 'exit':
```
gatesn, line 104: What is this for? Why does the server know about a 'exit' method?

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+            return
+
+        params = request.params if request.params is not None else {}
```
gatesn, line 107: `params = request.params or {}`

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+            maybe_handler = self._message_handler(request.method, params)
+        except KeyError:
+            log.debug("No handler found for %s", request.method)
+            self._message_manager.write_message(
+                JSONRPC20Response(_id=request._id, error=JSONRPCMethodNotFound()._data).data)
+            return
+
+        if request._id in self._received_requests:
+            log.error('Received request %s with duplicate id', request.data)
+        elif callable(maybe_handler):
```
gatesn, line 118: I think this might be the wrong place to check callable. How can I have two linting hooks, one sync and one async right now?

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+        future = self._executor_service.submit(handler)
+
+        if request.is_notification:
+            return
+
+        self._received_requests[request._id] = future
+
+        def did_finish_callback(completed_future):
+            if completed_future.cancelled():
+                log.debug('Cleared cancelled request %d', request._id)
```
gatesn: `s/Cleared/Clearing/`

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+                del self._received_requests[request._id]
+                return
+
+            error, trace = completed_future.exception_info()
+            del self._received_requests[request._id]
```
gatesn: Why this line here, and above in the if statement.

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+            if error is not None:
+                log.error("Failed to handle request %s with error %s %s", request._id, error, trace)
+                # TODO(forozco): add more descriptive error
+                response = _make_response(request, errror=JSONRPCInternalError()._data)
```
gatesn: `s/errror/error/`

pyls/rpc_manager.py
```diff
@@ -0,0 +1,171 @@
...
+            else:
+                response = _make_response(request, result=completed_future.result())
+            self._message_manager.write_message(response.data)
+        future.add_done_callback(did_finish_callback)
+
+    def _handle_response(self, response):
+        try:
+            request = self._sent_requests[response._id]
```
gatesn, line 151: Just guard with `if response._id not in self._sent_requests` - cleaner to read

pyls/workspace.py
```diff
@@ -74,12 +74,12 @@ class Workspace(object):
     M_SHOW_MESSAGE = 'window/showMessage'
     PRELOADED_MODULES = get_preferred_submodules()
 
-    def __init__(self, root_uri, lang_server=None):
+    def __init__(self, root_uri, rpc_manager=None):
```
gatesn, line 77: Don't like that everything is a manager.

test/fixtures.py
```diff
@@ -1,10 +1,11 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import pytest
+import os
 from pyls import uris
 from pyls.config.config import Config
 from pyls.python_ls import PythonLanguageServer
 from pyls.workspace import Workspace
-from io import StringIO
+from StringIO import StringIO
```
gatesn, line 8: Revert this: 

```
io.StringIO is a class. It handles Unicode. It reflects the preferred Python 3 library structure.

StringIO.StringIO is a class. It handles strings. It reflects the legacy Python 2 library structure.
```

test/fixtures.py
```diff
@@ -1,10 +1,11 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import pytest
+import os
```
gatesn, line 3: Unused?

test/test_rpc_manager.py
```diff
@@ -0,0 +1,123 @@
+# Copyright 2018 Palantir Technologies, Inc.
+import pytest
+import time
+from StringIO import StringIO
```
gatesn, line 4: `io.StringIO`

gatesn (commented):
Haven't reviewed tests, some nice looking stuff here. One thing about the async is that there's no way (I think?) to mix async and sync hook implementations. 

This might end up quite a large refactor to support this...

pyls/message_manager.py
ferozco: no, had thought about implementation but forgot to add it. can do it here or in a follow up

pyls/message_manager.py
ferozco: only need to do this to handle to weirdness within JSONRPC20Response constructor, JSONRPCRequest handles things properly

pyls/message_manager.py
ferozco: no we should not. consider the case that shutdown has been received and processed but there were some pending handlers that had not finished executing. When those handlers finish, rather than throwing and crashing we quietly drop the messages until we receive the exit notification and cleanup properly

pyls/python_ls.py
ferozco: oh oops, not sure how that works haha

pyls/python_ls.py
ferozco: i don't think it really matters

pyls/python_ls.py
ferozco: actually did this explicitly to avoid collision with `id` keyword

pyls/rpc_manager.py
ferozco: I wouldn't personally call either of the managers servers. neither really provide server like functionality to the client, they are mostly just there to manage the sending of messages so that the real server can more easily operate

pyls/rpc_manager.py
ferozco: Thats actually not true. This doesn't let you operate on requests send from the language server to the client(whether we should is a totally separate question). The reason for having this method is to allow the language server correctly act upon `$/cancel` requests send from the client

pyls/rpc_manager.py
ferozco: I suppose the language server can `drop` the message within the handler, but it complicates the logic since this method now has to know how to differentiate between an empty response and a response that should be dropped

pyls/rpc_manager.py
ferozco: I think being more explicit is fine

pyls/rpc_manager.py
ferozco: Linting is a notification so I don't think it wouldn't be effected by this. Perhaps i didn't mention this, but the plan is to have the server do the resolution around having some async and some non-async hooks. i.e. for each handler there would be some merge logic to determine whether to return an aggregate result or a callable

pyls/rpc_manager.py
ferozco: I think the current language is consistent with the other log lines

pyls/rpc_manager.py
ferozco: my bad

pyls/rpc_manager.py
ferozco: its more pythonic to do it this way, EAFP

pyls/workspace.py
ferozco: we can talk about this more offline

test/fixtures.py
ferozco: ya thats something i noticed, didn't seem like linting was actually being run against tests

test/test_rpc_manager.py
ferozco: StringIO isn't the right way, but neither is io.StringIO. it forces you to have to convert all your strings before sending them off, which is totally unnecessary during normal execution

evandrocoan: This will not run on Python 3 and this server is compatible is both Python 2 and 3. You need to:
```python
import sys

if sys.version_info[0] < 3:
    from StringIO import StringIO

else:
    from io import StringIO
```

pyls/rpc_manager.py
evandrocoan: Whether you do `LBYL` or `EAFP` is only relevant for performance reasons. If this is supposed to fail 99% of the time, do `LBYL` because exception handingly 99% of the time is much more extensive than the if checking.

Anyways, only optimize after you know what needs optimization, i.e., after benchmarking.

> In DonaldKnuth's paper "StructuredProgrammingWithGoToStatements", he wrote: "Programmers waste enormous amounts of time thinking about, or worrying about, the speed of noncritical parts of their programs, and these attempts at efficiency actually have a strong negative impact when debugging and maintenance are considered. We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil. Yet we should not pass up our opportunities in that critical 3%." http://wiki.c2.com/?PrematureOptimization

pyls/rpc_manager.py
evandrocoan: It is just `Microsoft Language Server Protocol`. VSCode just happens to be one of the first to use it.

ferozco:
@evandrocoan thanks for looking at this! unfortunately this PR was closed in favour of https://github.com/palantir/python-language-server/pull/261

PR #266, opened 2018-02-21
Fix allowing plugins to disable themselves by default

gatesn:


pyls/config/config.py
```diff
@@ -64,6 +65,8 @@ def settings(self, document_path=None):
             2. Plugin settings, reported by PyLS plugins
             3. LSP settings, given to us from didChangeConfiguration
             4. Project settings, found in config files in the current project.
+
+        TODO(gatesn): We should update a cached view of this whenever any configs change
```
ferozco, line 69: do you think we should memoize this? seems a little extra to reconstruct the settings each time

ferozco (approved):

PR #276, opened 2018-02-26
update install requires

ferozco:
Ensure futures backport is pulled in correctly


gatesn:
Why less than 3.2? Why not less than 3?

ferozco:
it was added in 3.2: https://docs.python.org/3/library/concurrent.futures.html

gatesn (approved):

PR #278, opened 2018-02-26
Strip line with \r\n before calculation of max column

danielnowakowski:
When cursor is in the last line, at the end of the line and user requests for completions, column position is calculated incorrectly. It happens due to missing new line character. In this fix, line is stripped before calculation of its length and received value is taken as max_column.


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @danielnowakowski! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn (approved):
Looks great! Thanks for your contribution

PR #279, opened 2018-02-26
introduce handling of JEDI_SYS_PATH

danielnowakowski:
Introduction of a new environment variable JEDI_SYS_PATH. When this variable is set, it is added to path passing to Jedi as sys_path.

This feature is being added to support Jedi execution within another sys.path.

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -258,8 +258,12 @@ class Document(object):
         # Copy our extra sys path
         path = list(self._extra_sys_path)
 
+        # Check if JEDI_SYS_PATH is available and use it if yes
+        if 'JEDI_SYS_PATH' in os.environ:
+            log.info("Using JEDI_SYS_PATH %s", os.environ['JEDI_SYS_PATH'])
+            path.extend(os.environ['JEDI_SYS_PATH'].split(os.pathsep))
         # Check to see if we're in a virtualenv
-        if 'VIRTUAL_ENV' in os.environ:
+        elif 'VIRTUAL_ENV' in os.environ:
             log.info("Using virtualenv %s", os.environ['VIRTUAL_ENV'])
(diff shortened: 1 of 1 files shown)


pyls/workspace.py
```diff
@@ -258,8 +258,12 @@ def sys_path(self):
         # Copy our extra sys path
         path = list(self._extra_sys_path)
 
+        # Check if JEDI_SYS_PATH is available and use it if yes
+        if 'JEDI_SYS_PATH' in os.environ:
+            log.info("Using JEDI_SYS_PATH %s", os.environ['JEDI_SYS_PATH'])
+            path.extend(os.environ['JEDI_SYS_PATH'].split(os.pathsep))
         # Check to see if we're in a virtualenv
-        if 'VIRTUAL_ENV' in os.environ:
+        elif 'VIRTUAL_ENV' in os.environ:
```
gatesn, line 266: Why elif? Why not just keep extending? And in that case, JEDI probably wants to take priority over your VENV.

gatesn:
It doesn't seem like JEDI_SYS_PATH is a well-used thing. What are you trying to achieve here?

danielnowakowski:
Let's consider the situation, where you have completely different working directory with another Python version and another packages. And you want to run Jedi with path(s) pointing to that directories. Thank to my change, you can set JEDI_SYS_PATH and tell JEDI where to look for modules.

gatesn:
I think Jedi might be too low level here. And we might just want a config options for "extraSysPath". This can then be read by all extensions not just Jedi

gatesn:
Closing this out. A preferred solution would be to specify an environment or sys path for Jedi to use. Jedi 0.12.0 includes a new API for environment handling, perhaps we could make that configurable?

PR #288, opened 2018-03-14
debounce based off of inputs

### LINKED ISSUE #285: Diagnostics calculations
Let's say you have client with two open files and after entering into your client, you perform initialization and later you send didOpen notifications for these two files. You expect that diagnostics will be calculated for both of the files, however result is got only for one file. It probably happens due to lint deboucing, which is provided in the implementation of language server. 

This behavior seems to be incorrect.

In my opinion debouncing should be considered only for the same file.

ferozco:
key the debouncing off of the value of the inputs. Closes #285


pyls/_utils.py
```diff
@@ -9,21 +9,49 @@
...
 
-def debounce(interval_s):
+def debounce(interval_s, keys=None):
     """Debounce calls to this function until interval_s seconds have passed."""
     def wrapper(func):
+        arg_cache = {}
+        lock = threading.Lock()
+
+        @functools.wraps(func)
+        def cleanup(*args, **kwargs):
```
gatesn: cleanup_and_run

pyls/_utils.py
```diff
@@ -9,21 +9,49 @@
...
 FIRST_CAP_RE = re.compile('(.)([A-Z][a-z]+)')
 ALL_CAP_RE = re.compile('([a-z0-9])([A-Z])')
+KWD_MARK = object()
 
 
-def debounce(interval_s):
+def debounce(interval_s, keys=None):
     """Debounce calls to this function until interval_s seconds have passed."""
     def wrapper(func):
+        arg_cache = {}
```
gatesn: timers

pyls/_utils.py
```diff
@@ -9,21 +9,49 @@
...
+        lock = threading.Lock()
+
+        @functools.wraps(func)
+        def cleanup(*args, **kwargs):
+            input_hash = _hash_input(keys, *args, **kwargs)
+            with lock:
+                del arg_cache[input_hash]
+            return func(*args, **kwargs)
+
         @functools.wraps(func)
```
gatesn: Remove the `wraps` decoration

pyls/_utils.py
```diff
@@ -9,21 +9,49 @@
...
-def debounce(interval_s):
+def debounce(interval_s, keys=None):
     """Debounce calls to this function until interval_s seconds have passed."""
     def wrapper(func):
+        arg_cache = {}
+        lock = threading.Lock()
+
+        @functools.wraps(func)
+        def cleanup(*args, **kwargs):
+            input_hash = _hash_input(keys, *args, **kwargs)
```
gatesn: Hoist computing the input_hash up to where you create the lock

pyls/_utils.py
```diff
@@ -9,21 +9,49 @@
...
+                if input_hash in arg_cache:
+                    arg_cache[input_hash].cancel()
+                timer = threading.Timer(interval_s, cleanup, args, kwargs)
+                arg_cache[input_hash] = timer
+                timer.start()
         return debounced
     return wrapper
 
 
+def _hash_input(keys, *args, **kwargs):
```
gatesn: Get rid of this and just use `kwargs[key]` - can make it more complicate if/when you need it

gatesn (changes requested):

pyls/_utils.py
ferozco: you can't do that! the input_hash depends on the input and has to be recomputed each time the function is called. meanwhile the lock is just a member of the debounce closure.

gatesn: Oh yup, you're right. You could pass the key to `cleanup` as an argument. But not sure you gain much

pyls/_utils.py
gatesn: Actually, this is wrong. We should leave it on debounced and remove it from cleanup. The decorator basically copies the docstrings / other properties over. No point having it on cleanup. It's debounced that we return

gatesn (approved):

PR #289, opened 2018-03-14
handle cancel correctly

ferozco:
Per LSP spec (AKA bible) every request must be responded to, even cancelled ones. We now do that correctly.

### DIFF
--- a/pyls/rpc_manager.py
+++ b/pyls/rpc_manager.py
@@ -84,7 +84,8 @@ class JSONRPCManager(object):
         log.debug('Cancel request %d', request_id)
         try:
             self._received_requests[request_id].set_exception(
-                JSONRPCDispatchException(code=LSP_CANCEL_CODE, message="Request cancelled"))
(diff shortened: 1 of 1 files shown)


pyls/rpc_manager.py
```diff
@@ -82,7 +83,8 @@ def cancel(self, request_id):
         """
         log.debug('Cancel request %d', request_id)
         try:
-            self._received_requests[request_id].cancel()
+            self._received_requests[request_id].set_exception(
+                JSONRPCDispatchException(code=LSP_CANCEL_CODE, message="Request cancelled"))
```
gatesn: Hanging parenthesis!

gatesn: And can you put the request id in the message

ferozco: don't need the request id! the actual response is create here: https://github.com/palantir/python-language-server/blob/develop/pyls/rpc_manager.py#L150

gatesn (approved):

PR #298, opened 2018-03-22
Lazily construct rope projects and resources

### LINKED ISSUE #179: Make rope/jedi module pre-loading configurable
Right now there's a statically defined list of modules that we try to eagerly import. At the cost of memory, this can drastically reduce the time-to-first-completion.

### LINKED ISSUE #256: Load the server plugins after receiving the language client configurations
Otherwise we cannot pass setup settings to the server plugins:

1. https://github.com/palantir/python-language-server/pull/253#discussion_r164610715 Created the rope.create_folder setting to not create the
1. https://github.com/palantir/python-language-server/issues/179 Make rope/jedi module pre-loading configurable

gatesn:
To allow user settings to affect rope.

Fixes #179 
Fixes #253 
Fixes #256


pyls/workspace.py
```diff
@@ -159,14 +163,13 @@ def __init__(self, uri, source=None, version=None, local=True, extra_sys_path=No
...
-        self._rope_project = rope
+        self._rope_project_builder = rope_project_builder
 
     def __str__(self):
         return str(self.uri)
 
-    @property
-    def _rope(self):
-        return libutils.path_to_resource(self._rope_project, self.path)
+    def _rope_resource(self, rope_config):
```
ferozco, line 171: why is this a `private method` if it is only used outside of the class?

ferozco (approved):

pyls/workspace.py
gatesn: Sort of abused the "private" convention to keep the rope APIs hidden. I consider them experimental / unstable and might even be removed in the future.

PR #310, opened 2018-03-26
Populate PRELOADED_MODULES lazily

nanoant:
Prevents wasting CPU/IO cycles when PRELOADED_MODULES are not going to be
used anyway, e.g. Rope is missing or disabled.

This is follow-up for #286 and #301

PR #311, opened 2018-03-26
configparser is built-in in Python>=3.0

nanoant:
If I understand it correctly we need `configparser` only for <3.0 (2.7), because `configparser` is built-in for `>="3.0"` and we don't use it explicitly as a backport.

Quoting: https://pypi.python.org/pypi/configparser
> To use the configparser backport instead of the built-in version on both Python 2 and Python 3, simply import it explicitly as a backport: `from backports import configparser`
> If you’d like to use the backport on Python 2 and the built-in version on Python 3, use that invocation instead: `import configparser`

And the 2nd option `import configparser` is what is used in https://github.com/palantir/python-language-server/blob/develop/pyls/config/source.py, so how about making this dependency `python_version<"3.0"` only?

Additional NOTE for Conda users: `configparser` is available only for Python 2.7, so this dependency cannot be satisfied with Conda and Python 3.x.

~~~
> conda search configparser
Loading channels: done
# Name                  Version           Build  Channel
configparser            3.5.0b2          py27_1  pkgs/free
configparser              3.5.0          py27_0  pkgs/free
configparser              3.5.0  py27h2fa79a8_0  pkgs/main
~~~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
     # requirements files see:
     # https://packaging.python.org/en/latest/requirements.html
     install_requires=[
-        'configparser',
+        'configparser; python_version<"3.0"',
         'future>=0.14.0',
         'futures; python_version<"3.2"',
         'jedi>=0.10',


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @nanoant! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nanoant:
Sorry for a noise, I realized too late that I don't really want to sign a CLA & copyright transfer for such a little change 😞

gatesn:
@nanoant please do contribute PRs anyway. I am happy to re-apply the patch under my name if it's appropriate. As per #316

nanoant:
@gatesn Okay, gimme 1 day for clarify on CLA. I think I should be good to go then.

nanoant:
@gatesn FYI I am okay with CLA, so I have reopened my other PR #310. Thanks a lot for your support.

PR #336, opened 2018-04-17
Support underscore in pycodestyle config

maxnordlund:
Pycodestyle also supports using underscore as the separator in multi word
options, and pyls should as well.

### DIFF
--- a/pyls/config/pycodestyle_conf.py
+++ b/pyls/config/pycodestyle_conf.py
@@ -12,8 +12,10 @@ OPTIONS = [
     ('exclude', 'plugins.pycodestyle.exclude', list),
     ('filename', 'plugins.pycodestyle.filename', list),
     ('hang-closing', 'plugins.pycodestyle.hangClosing', bool),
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @maxnordlund! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn:
Doesn’t this line handle the case? https://github.com/palantir/python-language-server/blob/develop/pyls/config/source.py#L62

maxnordlund:
What do you know, I didn't see that. If it does not this can always be redone.

gatesn:
Only asking in case this was driven by a bug you found, in which case we should verify it works / write a test.

maxnordlund:
I'm not totally sure, but I did noticed that my configs weren't picked up properly and the only difference seemed to be underscores vs dashes.

After changing my configs to use dashes it started to work, then I made this PR and kinda forgot about it. Now that it works again I mean.

PR #344, opened 2018-05-04
Fix running with -m

### LINKED ISSUE #258: Running pyls with python -m pyls
Thank you for the superb package!

Is there a way to specify which version of python you want to run `pyls` under?

I've tried running `python2 -m pyls`, which is what I do for pytest, but that didn't seem to work

gatesn:
Fixes #258 

When running with `-m` this means `pyls` runs the main function


gatesn:
Seems like we have a few circular dependencies. Plus the urllib.parse issue is a bit odd. Seems like futures might have changed their behaviour

gatesn:
Ah, I'd forgotten to add `if __name__ == "__main__"` to the main script

Issue #218, opened 2018-01-05
Cannot LspRename


mcepl:
I have filed https://github.com/prabirshrestha/vim-lsp/issues/61 but the maintainer there (@prabirshrestha) closed the bug on me blaming the server. Even when I upgraded python-language-server to the last release (0.12.1) I get the same reaction (this is the [log](https://github.com/palantir/python-language-server/files/1606376/vim-lsp.log.txt)): no error, but no change in the editor.

Using vim 8, python 3.4, on RHEL-7.

Any ideas, what's wrong?

gatesn:
Are you able to get me the stderr of the python language server? Seems like the log you linked just includes the vim-lsp logs.

prabirshrestha:
@mcepl Can you update the log with verbose to enable stderr messages. `let g:lsp_log_verbose = 1`
@gatesn For me even log verbose didn't show any messages. Is there a way to force pyls to use its own log file?

mcepl:
> Can you update the log with verbose to enable stderr messages. let g:lsp_log_verbose = 1

I have ``let g:lsp_log_verbose = 1`` in my ``~/.vim/vimrc``.

So, I tried to add this to the configuration in ``~/.vim/vimrc``:

```viml
if executable('pyls')
    " pip install python-language-server
    au User lsp_setup call lsp#register_server({
        \ 'name': 'pyls',
        \ 'cmd': {server_info->['pyls', '--verbose', '--log-file', '/tmp/pyls-log.txt']},
        \ 'whitelist': ['python'],
        \ })
endif
```
And attached is both [pyls-log.txt](https://github.com/palantir/python-language-server/files/1618144/pyls-log.txt) and [vim-lsp.log](https://github.com/palantir/python-language-server/files/1618147/vim-lsp.log).

gatesn:
@mcepl could you try `pyls -vv --log-file ...` that should get debug messages from pyls.

mcepl:
[pyls-log.txt](https://github.com/palantir/python-language-server/files/1618432/pyls-log.txt) and [vim-lsp.log](https://github.com/palantir/python-language-server/files/1618433/vim-lsp.log)

gatesn:
That's really weird, the pyls log just stops abruptly, and the vim log claims python exited 1. I'll try and reproduce

mcepl:
Still doesn’t work even with python-language-server 0.14.0, latest vim-lsp and vim 8.0.1520.

[...]

PR #345, opened 2018-05-07
rename now reports correct number of lines.

abingham:
It was initially reporting `sys.maxsize` all of the time, and this was both
incorrect and problematic for some clients.

See #218 for more information.

### DIFF
--- a/pyls/plugins/rope_rename.py
+++ b/pyls/plugins/rope_rename.py
@@ -1,9 +1,6 @@
 # Copyright 2017 Palantir Technologies, Inc.
-from functools import reduce
-import io
 import logging
 import os
-import sys
 
 from rope.base import libutils
 from rope.refactor.rename import Rename
@@ -11,13 +8,6 @@ from rope.refactor.rename import Rename
 from pyls import hookimpl, uris
 
 log = logging.getLogger(__name__)
-
-
-def _num_lines(resource):
-    """Count the number of lines in a `File` resource.
-    """
-    s = io.StringIO(resource.read())
-    return reduce(lambda x, _: x + 1, s, 0)
 
 
 @hookimpl
@@ -50,3 +40,8 @@ def pyls_rename(config, workspace, document, position, new_name):
             }]
         } for change in changeset.changes]
     }
+
+
+def _num_lines(resource):
+    "Count the number of lines in a `File` resource."
+    return len(resource.read().splitlines())
--- a/pyls/plugins/rope_rename.py
+++ b/pyls/plugins/rope_rename.py
@@ -1,4 +1,6 @@
 # Copyright 2017 Palantir Technologies, Inc.
+from functools import reduce
+import io
 import logging
 import os
 import sys
@@ -9,6 +11,13 @@ from rope.refactor.rename import Rename
 from pyls import hookimpl, uris
 
 log = logging.getLogger(__name__)
+
+
+def _num_lines(resource):
+    """Count the number of lines in a `File` resource.
+    """
+    s = io.StringIO(resource.read())
+    return reduce(lambda x, _: x + 1, s, 0)
 
 
 @hookimpl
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @abingham! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pyls/plugins/rope_rename.py
```diff
@@ -1,4 +1,6 @@
 # Copyright 2017 Palantir Technologies, Inc.
+from functools import reduce
```
gatesn: Can you just import functools? I (and probably the linter!) don’t like overriding the built-in function.

pyls/plugins/rope_rename.py
```diff
@@ -11,6 +13,13 @@
 log = logging.getLogger(__name__)
 
 
+def _num_lines(resource):
```
gatesn: Can you move “private” functions to bottom of module

pyls/plugins/rope_rename.py
```diff
@@ -11,6 +13,13 @@
...
+    """Count the number of lines in a `File` resource.
+    """
```
gatesn: Closing quotes on the same line if they fit

pyls/plugins/rope_rename.py
```diff
@@ -11,6 +13,13 @@
...
+    s = io.StringIO(resource.read())
```
gatesn: What does resource.read() return? Seems unnecessary to use reduce given you’re loading the resource into memory anyway. Can’t you just call splitlines and count it?

gatesn (commented):
Some style comments

pyls/plugins/rope_rename.py
abingham: Yeah, the `reduce` call was a bit too cutesy.

abingham:
I think I've made all the changes you asked for. I also added `version` to the `textDocument` response; I did this because the emacs client I'm using requires it, and this is how I was able to test the change. It also seems required by LSP, so I figured it was OK.

pyls/plugins/rope_rename.py
```diff
@@ -31,13 +30,19 @@ def pyls_rename(config, workspace, document, position, new_name):
                 'uri': uris.uri_with(
                     document.uri, path=os.path.join(workspace.root_path, change.resource.path)
                 ),
+                'version': document.version
```
gatesn: This is incorrect, the `uri` we're returning is derived from the rope resource (line above), however this version number is derived from the document within which the rename was triggered.

If you check the LSP, version is not a required field: https://microsoft.github.io/language-server-protocol/specification#versionedtextdocumentidentifier

However, we can do better than just never setting it. If you use `workspace.get_document(doc_uri)` and then use that document's version then it will return None (correctly) if the server has never received a didOpen message, or the correct version number if it has. 

My concern with doing this is that rope doesn't operate on the in-memory view of the documents. We should probably fix that before making too heavy usage of rope. In particular, the edits returned by rope could be inconsistent with the version of the document.

gatesn (changes requested):

pyls/plugins/rope_rename.py
abingham: So is it incorrect that the client is requiring the version? I think they're only accidentally requiring it, but I don't know for sure. 

Regarding rope: yes, it would be nice if we could get it to use the in-memory buffer. In my earlier work with rope I just had the client save the file, but that was the lazy way out.

gatesn: Yup, I think it's incorrect of the client to require it. Though it might be correct if they only require it for documents that the client has sent a didOpen message for

abingham:
The `lsp-mode` emacs client code now behaves correctly in the face of a missing version, and I think this PR has all of the changes you asked for.

PR #347, opened 2018-05-09
Document highlights

### LINKED ISSUE #227: Support document highlights
Using Jedi we can extract read/write uses of a symbol.

gatesn:
Fixes #227


lilifan:
Sir, is syntax highlighting enabled out of box?  or could you tell me how to install and enable it.  My first try shows no highlights. All words are same color.  @gatesn

PR #348, opened 2018-05-10
Basic RST -> Markdown

### LINKED ISSUE #291: pydoc rendering on datatip at ide-python with atom-ide-ui
37638320-2d413d24-2c47-11e8-8405-74d4654665c8

It looks that the pydoc was rendered as markdown.
Can you fix it?

Owner of ide-python said:
>I'm afraid this requires some upstream fixes in either the python language server or the protocol specifications.
The problem is that the LSP currently only supports markdown or plain text. Since Python docstrings are usually written in rst the language server would need to convert it to markdown first. Ideally this could also be handled by the frontend but this also requires (small) changes to the protocol itself.

Also, here are issues related to this problem:
https://github.com/lgeiger/ide-python/issues/95
https://github.com/facebook-atom/atom-ide-ui/issues/23

gatesn:
Fixes #291


pyls/plugins/hover.py
```diff
@@ -14,7 +14,10 @@ def pyls_hover(document, position):
...
 
     if not definitions:
-        # :(
-        return {'contents': ''}
+        return None
 
-    return {'contents': _utils.format_docstring(definitions[0].docstring()) or ""}
+    md_docstring = _utils.format_docstring(definitions[0].docstring())
+    return {'contents': {
+        'type': lsp.MarkupKind.Markdown,
```
lgeiger: According to the [spec](https://microsoft.github.io/language-server-protocol/specification) the this should be `'kind': lsp.MarkupKind.Markdown` instead of `'type': lsp.MarkupKind.Markdown` since hover requests return a `MarkupContent` object:

```ts
export interface MarkupContent {
	/**
	 * The type of the Markup
	 */
	kind: MarkupKind;

	/**
	 * The content itself
	 */
	value: string;
}
```

lgeiger (commented):
Thanks @gatesn for getting this started. This is a huge improvement 💯

gatesn:
And that’ll be why I’ve seen no marked improvement in VSCode!! Thanks for pointing that out.

I also note that docutils exists in Python, so we might want to make use of its RST parser instead of replicating the jacket typescript one in the VSCode plugin.

PR #355, opened 2018-05-23
fix config bug

purpleP:
I suppose it's a bug, but I'm not sure, because I don't fully understand from where are all possible configurations can be taken.

### DIFF
--- a/pyls/config/config.py
+++ b/pyls/config/config.py
@@ -91,7 +91,9 @@ class Config(object):
         sources = self._settings.get('configurationSources', DEFAULT_CONFIG_SOURCES)
 
         for source_name in reversed(sources):
-            source = self._config_sources[source_name]
+            source = self._config_sources.get(source_name)
+            if not source:
+                continue
             source_conf = source.user_config()
             log.debug("Got user config from %s: %s", source.__class__.__name__, source_conf)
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @purpleP! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn:
Yup, was introduced in this PR: https://github.com/palantir/python-language-server/pull/335/files

Thanks!

gatesn (approved):

purpleP:
@gatesn I agree about `try expect continue` being more pythonic here, but I should add that there's a bigger problem here - code duplication. I can refactor that if someone would explain to me how this supposed to work. Seems like maybe one could change `merge_dicts` to take variable number of dicts instead of just two and just merge all configs into one.

gatesn:
Yeah, agree this could be refactored to be much nicer. Config providers should be lazily loaded iff they are installed, merging configs should be cached etc. Happy for you to take a look at this if you like? Either way, we should get this bug fix in I think.

PR #356, opened 2018-05-23
use difflib to generate TextEdits

purpleP:
According to the language server specification `range.end` is exclusive, but neovim languge client seems to ignore that and when I try to format current line it sends the same line as start and end instead of `{"start": {"line": current_line, "character": 0}, "end": {"line": current_line + 1, "character": 0}}`. That's why I added

`end + (start == end)`

### DIFF
--- a/pyls/plugins/yapf_format.py
+++ b/pyls/plugins/yapf_format.py
@@ -1,6 +1,11 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import logging
 import os
+
+from bisect import bisect
+from difflib import SequenceMatcher
+from itertools import accumulate
+
 from yapf.yapflib import file_resources
 from yapf.yapflib.yapf_api import FormatCode
 from pyls import hookimpl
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @purpleP! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

DerWeh:
Correct me if I am wrong, but I think you use `get` only to avoid invalid keys? Are there really config sources that are `None`? 

If you only check for invalid keys, the pythonic way should be 

```python
try:
    source = self._config_source[source_name]
except KeyError:
    continue
```

This avoids unnecessary `if`s, it is easier to ask for forgiveness than permission.

gatesn:
@DerWeh think you meant to comment on the other PR!

purpleP:
@DerWeh, yeah, `try except` is probably more pythonic, although it takes one more line. I used `get` because initially I thought about using default value instead of `continue`. This needs to be refactored anyway, because there's a code duplication.

PR #379, opened 2018-06-15
highlight should return an empty list instead of None

pixystone:
`pyls_document_highlight` would return an empty list when no usages found.

It is unnecessary that changing `[] or None` to `None`, just return `[]` instead.

This would be more compatible for various kinds of language clients.

### DIFF
--- a/pyls/python_ls.py
+++ b/pyls/python_ls.py
@@ -187,7 +187,7 @@ class PythonLanguageServer(MethodDispatcher):
         return self._hook('pyls_format_range', doc_uri, range=range)
 
     def highlight(self, doc_uri, position):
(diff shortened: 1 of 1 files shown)


gatesn:
Which language clients can't handle this? It's part of the spec:

```
Response:
result: DocumentHighlight[] | null defined as follows:
```

pixystone:
Somebody write a client and no longer maintain it now. It is hard to find the source code, so maybe I could make server more compatible.

Just keep the same with what the`pyls_document_highlight` returns.

gatesn:
I don't think we want to make a habit of writing our servers to conform to unmaintained clients. I'm going to close this out since we comply with the LSP in this regard.

PR #382, opened 2018-06-21
Added pyls.extraSysPath. Fixes #381

### LINKED ISSUE #381: Allow passing sys_path to Jedi from the settings
Additionally to `source_roots`, it should be possible to pass additional `sys_path` from the settings. Issues arise when using Python2/Python3 codebases and simply setting `PYTHONPATH`.

Kronuz:
This adds `extraSysPath` to extend `extra_sys_path` passed to Jedi.

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -21,9 +21,10 @@ class Workspace(object):
     M_APPLY_EDIT = 'workspace/applyEdit'
     M_SHOW_MESSAGE = 'window/showMessage'
 
-    def __init__(self, root_uri, endpoint):
+    def __init__(self, root_uri, endpoint, config):
         self._root_uri = root_uri
         self._endpoint = endpoint
+        self._config = config
         self._root_uri_scheme = uris.urlparse(self._root_uri)[0]
         self._root_path = uris.to_fs_path(self._root_uri)
         self._docs = {}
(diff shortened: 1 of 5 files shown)


pyls/workspace.py
```diff
@@ -93,9 +94,12 @@ def source_roots(self, document_path):
 
     def _create_document(self, doc_uri, source=None, version=None):
         path = uris.to_fs_path(doc_uri)
+        extra_sys_path = self.source_roots(path)
+        settings = self._config.settings(doc_uri)
+        extra_sys_path.extend(settings.get('extraSysPath', []))
```
gatesn, line 99: Extend? Or prepend? I wonder if an explicit config override should be put at the front of the sys path

gatesn (changes requested):
Think we should have `settings.get('extraSysPath', []) + self.source_roots(path)`

gagbo:
Should the work on this start again, would this be merged ? I don't understand why it stalled exactly, and this is exactly the feature I'm missing.

I think I prefer prepending the extra_sys_path (or maybe give one after and one before options), because I would want to put specific `PYTHONPATH` values here, which are inserted early in `sys.path()` when running a python program. Especially when, as @gatesn says, it is an override.

**EDIT** : Actually, setting PYTHONPATH correctly should do the trick already (from adding a few logs in 0.26 code). At least, `sys_path` has a correct value, but jedi still doesn't pick up the folder like I want. I'll make a minimal example soon hopefully

ccordoba12:
This was implemented in PR #680.

PR #384, opened 2018-06-26
Add pylint support

cbosdo:
While testing onivim and python support to hack on salt, I discovered that there is no pylint support for the python-language-server. As hacking on salt requires it, I just added it.

I also fixed a line number bug in the mccabe plugin


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @cbosdo! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn (commented):
Thanks for your contribution! However, the reason this hasn’t been done before is because there’s no obvious way of having pylint run against the in-memory documents. 

This linter will return results for the contents on disk, regardless of what the user has typed but not yet saved. 

Given this isn’t likely to be resolved, perhaps we want to separate out onChange linting and onSave linting then only run pylint onSave?

cbosdo:
@gatesn hum... you're right. I guess that should be another PR? I'm not sure I'll have time soon to work on it. May be during [SUSE Hackweek](https://hackweek.suse.com/) 17 in early July

cbosdo:
@gatesn I think this fixes the problem you mentioned.

cbosdo:
considered didOpen like a didSave since the document isn't edited yet

gatesn:
Closed by #385

PR #392, opened 2018-07-07
Add mypy support using daemon

### LINKED ISSUE #391: Integrate with the mypy daemon
@gvanrossum creating a separate issue to discuss this.

I've poked around mypy a little and I'm very pleased at the sight of the fscache module. The reason pyls doesn't integrate with pylint is because we have no way of telling pylint to read from our workspace (in-memory view of what's in the client's buffers).

I imagine it would be pretty easy to patch the FileSystemCache class to read from our workspace. The downsides of this approach are:

* FileSystemCache isn't public API so it could change at any time. 
* We can't easily patch something inside the mypy daemon. And even if we can, we need to make it talk back to pyls.

Another approach might be to run the mypy daemon within the pyls process. The daemon's lifetime is then tied to that of your client which seems quite reasonable, particularly if we make use of the remote caching feature. I might give this a go.

gatesn:
Fixes #391 

One thing to note, I had to modify the mypy.errors module to write the traceback to stderr. Otherwise it breaks the LSP by writing to stdout.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,7 @@ ENV/
 
 # Merge orig files
 *.orig
+
+# Mypy
+.dmypy.json
+.mypy_cache/
--- a/pyls/plugins/mypy_lint.py
+++ b/pyls/plugins/mypy_lint.py
@@ -1,0 +1,157 @@
+# Copyright 2017 Palantir Technologies, Inc.
+import hashlib
+import logging
+import threading
+import re
+import sys
+import time
+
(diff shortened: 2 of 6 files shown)


eliwe:
From [follow-imports](http://mypy.readthedocs.io/en/latest/running_mypy.html#follow-imports) it seems the best option for error reporting is '--follow-imports silent'. Its probably expansive but as mypy is running in daemon mode it should not be that bad. Anything else will miss lots of errors.

PR #385, opened 2018-06-27
Add pylint support.

DanAlbert:
This also adds an `is_saved` argument to the pyls_lint hookspec, since
pylint doesn't expose an API for operating on in-memory contents, only
files.

Fixes #129

PR #393, opened 2018-07-18
Add Lint Trigger

drmathochist:
Pass a "trigger" string to the `pyls_lint` hook.  This allows linter
hook implementations to decide whether to run or not, based on what
triggered linting.  Useful for linters that are expensive to run,
so we don't want to run them on-change.

Since hook arguments are passed by keyword, this requires no changes
to existing local linter hook implementations.

### DIFF
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -88,7 +88,7 @@ def pyls_initialize(config, workspace):
 
 
 @hookspec
-def pyls_lint(config, workspace, document):
+def pyls_lint(config, workspace, document, trigger):
     pass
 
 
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @drmathochist! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn (changes requested):
Sorry it took me so long to get to this! Generally I like the approach, but I think the trigger strings should a) be called `triggered_by`, and b) be part of the hookspecs.py file in some sort of enum-style object (see the lsp.py file). 

That way they're more obviously part of the hook-spec API.

gatesn:
Going to close this since #385 adds a "saved" argument to the lint hook. For expensive linters moving forwards I suggest we add an async_lint hook that passes a callback function. This way the linter can take as long as it likes, and even call multiple times per request.

PR #404, opened 2018-08-20
Respect option `follow_imports` in `goto_assignments` call

st4lk:
Add an ability to specify `follow_imports`, `follow_builtin_imports` options in `goto_assignments` call: https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Script.goto_assignments

### DIFF
--- a/pyls/plugins/definition.py
+++ b/pyls/plugins/definition.py
@@ -7,8 +7,7 @@ log = logging.getLogger(__name__)
 
 @hookimpl
 def pyls_definitions(config, document, position):
-    expected_settings = {'follow_imports', 'follow_builtin_imports'}
-    params = {k: v for k, v in config.settings().items() if k in expected_settings}
+    params = {k: v for k, v in config.plugin_settings('jedi_definition').items() if k}
     definitions = document.jedi_script(position).goto_assignments(**params)
 
     definitions = [
--- a/pyls/plugins/definition.py
+++ b/pyls/plugins/definition.py
@@ -7,7 +7,7 @@ log = logging.getLogger(__name__)
 
 @hookimpl
 def pyls_definitions(config, document, position):
-    params = {k: v for k, v in config.plugin_settings('jedi_definition').items() if k}
+    params = {k: v for k, v in config.plugin_settings('jedi_definition').items() if v is not None}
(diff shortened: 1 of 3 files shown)


pyls/plugins/definition.py
```diff
@@ -6,8 +6,10 @@
 
 
 @hookimpl
-def pyls_definitions(document, position):
-    definitions = document.jedi_script(position).goto_assignments()
+def pyls_definitions(config, document, position):
+    expected_settings = {'follow_imports', 'follow_builtin_imports'}
```
gatesn: Don't tend to like defensively checking this.

pyls/plugins/definition.py
```diff
@@ -6,8 +6,10 @@
...
+    params = {k: v for k, v in config.settings().items() if k in expected_settings}
```
gatesn: Config should be namespaced by the plugin, e.g. `pyls.plugins.jedi_definition` https://github.com/palantir/python-language-server/blob/develop/vscode-client/package.json#L38

gatesn (changes requested):
Generally open to allowing more configuration here, but would like the config documented in package.json.

st4lk:
Updated

tmc:
@st4lk there are a couple of test failures: E       TypeError: pyls_definitions() takes exactly 3 arguments (2 given)

st4lk:
@tmc fixed

tmc:
@gatesn ping

pyls/plugins/definition.py
```diff
@@ -6,8 +6,9 @@
 
 
 @hookimpl
-def pyls_definitions(document, position):
-    definitions = document.jedi_script(position).goto_assignments()
+def pyls_definitions(config, document, position):
+    params = {k: v for k, v in config.plugin_settings('jedi_definition').items() if k}
```
gatesn: What does `if k` do here? Do you mean `if v` ? If you do, I think it should be `if v is not None`. Else if someone passes `False` for a property that jedi defaults to True, then the override won't be applied.

st4lk: @gatesn Fixed

gatesn (approved):

tmc:
can we get a release cut for this?

PR #410, opened 2018-08-28
Make python language server exit when parent process dies

jiahuijiang:


pyls/_utils.py
```diff
@@ -113,3 +113,19 @@ def clip_column(column, lines, line_number):
     # https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#position
     max_column = len(lines[line_number].rstrip('\r\n')) if len(lines) > line_number else 0
     return min(column, max_column)
+
+
+def is_process_alive(pid):
+    """ Check whether the process with the given pid is still alive.
+
+    :param pid: (int): process id
```
gatesn: Can you use Google-style docstrings instead?
```
Args:
    pid (int): process ID
```
```

File: pyls/_utils.py [hunk 2]
```diff
@@ -113,3 +113,19 @@ def clip_column(column, lines, line_number):
...
+    """
+    try:
+        os.kill(pid, 0)
+    except OSError as err:
+        # no such process or process is already dead
+        log.info("process %s is not alive: %s", pid, err)
```
gatesn: Bit agressive to log info this?

pyls/python_ls.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import logging
 import socketserver
+from threading import Thread, Timer
```
gatesn: Just `import threading`

pyls/_utils.py
jiahuijiang: 👍 moving the logging to ```watch_parent_process```

jiahuijiang:
@gatesn done :D

gatesn (approved):

PR #413, opened 2018-08-31
Check current folder for config source

st4lk:
If there is a config source (setup.cfg, .flake8, pycodestyle.cfg, etc) in the same folder as current document, I expect this document should be linted according to rules from this config.

Without this patch, such config is ignored.


gatesn:
@st4lk thank you for this PR, it's a good catch. I've opened another PR to apply the same patch to all plugins: https://github.com/palantir/python-language-server/pull/418

Issue #312, opened 2018-03-26
No errors returned from pydocstyle


andrewminer:
I've been using `pyls` (v0.16.0) by way of the `w0rp/ALE` (27c5f) plugin for Vim (v8.0 for MacOS) for some time now without any problems, but starting recently I'm not getting any errors from `pydocstyle`, even when I clearly have errors in my input file.  I've already been chatting with W0rp about it (w0rp/ALE#1407), and it doesn't look like there's anything wrong on that end of things.

To demonstrate the problem, I've created a very simple file with multiple kinds of problems:
``` python
class Foo(object):

    def __init__(self):
        self.bar = f"bar{baz}"
```

And captured a log file of `ALE`'s interactions with `pyls` while editing that file:

```
{"method":"textDocument/didChange","jsonrpc":"2.0","params":{"contentChanges":[{"text":"class Foo(object):\n\n    def __init__(self):\n        self.bar = f\"bar{baz}\"\n"}],"textDocument":{"uri":"file:///src/boom-mkboom/src/mkboom/sample.py","version":203}}}'
391.111639 SEND on 0: 'Content-Length: 130

{"method":"textDocument/didSave","jsonrpc":"2.0","params":{"textDocument":{"uri":"file:///src/boom-mkboom/src/mkboom/sample.py"}}}'
391.162486 : Starting job: sh  -c  git status --porcelain -- '/src/boom-mkboom/src/mkboom/sample.py'
391.162511 on 208: Created channel
391.179518 RECV on 208: '?? src/mkboom/sample.py
'
391.179734 on 208: channel_select_check(): Read EOF from ch_part[1], closing
391.191071 on 208: Job ended
391.191106 : Starting job: sh  -c  git status --porcelain -- '/src/boom-mkboom/src/mkboom/sample.py'
391.191136 on 209: Created channel
391.198092 : looking for messages on channels
391.198132 on 208: Closing channel because all readable fds are closed
391.198144 on 208: Closing channel
391.198152 on 208: Invoking callbacks and flushing buffers before closing
391.198225 on 208: Invoking channel callback <SNR>120_on_stdout
391.198317 on 208: Invoking close callback <SNR>120_on_exit_untracked
391.205380 RECV on 209: '?? src/mkboom/sample.py
'
391.205385 : looking for messages on channels
391.205399 on 209: Invoking channel callback <SNR>120_on_stdout
391.205595 on 209: channel_select_check(): Read EOF from ch_part[1], closing
[...]

PR #427, opened 2018-10-02
Explicitly state pydocstyle is disabled by default

rkiyanchuk:
Information in README is misleading listing `pydocstyle` among providers that will be enabled if found, as according to #312 it is disabled by default.

The README is updated to explicitly state that pydocstyle is disabled by default and minimal configuration to enable it is provided.


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @zoresvit! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn (approved):

PR #430, opened 2018-10-06
Clarify how to start the server

GeorgeBekh:


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @GeorgeBekh! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

README.rst
```diff
@@ -20,6 +20,10 @@ Symbols:
 
 ``pip install python-language-server``
 
+You can then start the server with this command:
+
+``python -m pyls``
```
gatesn, line 25: Does just running `pyls` not work for you?

GeorgeBekh: Doesn't seem like it. I install it without `sudo` so it probably shouldn't work in my case and in general

DerWeh: Did you make sure `pyls` is in your path?

GeorgeBekh: No, because i didn't know where it was located. Now i see that it's in `~/.local/bin/pyls` on my system and it's not in my `$PATH`.

I propose this change because it looks like this method (`python -m pyls`) is less ambiguous and error-prone. May be i'm wrong so please feel free to correct me

jgilmore: Less error-prone is right. I don't have ~/.local/bin/pyls on my computer. but python -m pyls works. pip show -v python-language-server enabled me to guess that it was in /usr/local/bin, which *is* in my path.

But I didn't know that command. And had to alias it to "python-language-server" for my editor plugin to find it.

gatesn:
I think we fixed this in https://github.com/palantir/python-language-server/pull/344

PR #443, opened 2018-10-22
setup.py: exclude all test packages

Cynerd:
There is a subpackage in test. This subpackage was before this commit
installed although test package was ignored. Extending exclude to ignore
anything that starts with 'test' is enough to exclude any subpackage of
test.

This is causing an error when trying to package pyls in Gentoo. Gentoo
build system disutils eclass checks if there is some test package to be
installed and fails installation step if so. This fixes that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setup(
         'configparser; python_version<"3.0"',
         'future>=0.14.0',
         'futures; python_version<"3.2"',
-        'jedi>=0.12',
+        'jedi>=0.13.2',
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @Cynerd! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #445, opened 2018-11-03
Fix parent process being killed on Windows when checking if it's alive

### LINKED ISSUE #434: Server crashes immediately with Neovim Language Client on Windows
Running `pip install python-language-server` and `pip install python-language-server[all]`

Then setting up dein in neovim like this:

```
call dein#add('Shougo/deoplete.nvim')

call dein#add('autozimu/LanguageClient-neovim')
```

And configuring language client like this:

```
let g:LanguageClient_serverCommands = {
    \ 'python': ['pyls', '-v'],
    \ }
```

Causes an immediate crash every time I open a python file.   I tried learning more about what is going on but couldn't get anywhere, not even a simple error message.

deathaxe:
Fixes #434

The `os.kill()` function being used to check for the existence of the parent process causes the owning process to be killed on Windows platforms.

This commit therefore introduces an alternative solution to check for the parent process to be alive by using the Windows API `types.windll.kernel32.OpenProcess()`

An alternative would have been to introduce psutil library as another dependency, which was judged not being worth.


rwols:
ping @gatesn @JiahuiJiang ?

ccordoba12:
I think this was wrongly closed.

jiahuijiang:
Right now if you run python language server without passing a --check-parent-process flag it will not check the parent process aliveness anymore

deathaxe:
... but if you do so on Windows, the parent process is still killed due to invalid usage of `os.kill()`.

This flag just asks the user to actively trigger the still existing bug, by passing the flag.

ccordoba12:
Yeah, that's why I think this should be reopened.

CAM-Gerlach:
This fix is critical on Windows; its apparently leading to many instances of zombie PyLS processes sticking around and taking up substantial memory and resources until the system is rebooted or the user knows to manually find and kill them. This and other stalled/rejected patches has forced us to maintain our own downstream fork, but we'd like to see these important fixes merged upstream to benefit all your users as well. Thanks!

ccordoba12:
> This and other stalled/rejected patches has forced us to maintain our own downstream fork

@gatesn, he's talking about Spyder (@CAM-Gerlach is one of our developers). We haven't decided yet if we would have to create a fork of this project for Spyder, but we will certainly have to if things like this PR are finally rejected.

Could you consider reviewing your decision? Thanks!

ccordoba12:
@andfoy, please create another PR based on this one and resubmit it to this repo so we can merge it. This basically makes the parent process check mechanism work on Windows.

ccordoba12:
@deathaxe, we have to that because tests are not passing here.

ccordoba12:
@deathaxe, or if you're still interested in merging this PR, please rebase it.

ccordoba12:
@deathaxe, @andfoy, please forget about my comments above. I forgot I could update this PR myself.

ccordoba12:
Thanks a lot @deathaxe for your contribution!

PR #460, opened 2018-12-09
Catch IndexError in pyls_signature_help

### LINKED ISSUE #459: IndexError in signature.py on textDocument/signatureHelp
I'm using pyls via [coc.nvim](https://github.com/neoclide/coc.nvim), and when it attempts to grab the signature of a class method, it's throwing an IndexError on this line:

https://github.com/palantir/python-language-server/blob/d7fb817e8f992bd5c6a287f5afb5e999a1e29449/pyls/plugins/signature.py#L24

It looks like in this case `s.docstring()` is somehow not returning a string, which causes `splitlines()` to not return an array, and `[0]` causes the error.

This was easy to mitigate by just catching that error and returning returning `.splitlines()` without the index, but I wanted to flag it here in case your team might have a better idea for why/when that wouldn't return an array, and what the right way to mitigate it is.

image

mike-hearn:
Fixes #459

This is how I'm addressing the issue locally, and it appears to fix the issue and causes the signature help to work correctly in coc.nvim.

### DIFF
--- a/pyls/plugins/signature.py
+++ b/pyls/plugins/signature.py
@@ -20,10 +20,16 @@ def pyls_signature_help(document, position):
         return {'signatures': []}
 
     s = signatures[0]
-    sig = {
-        'label': s.docstring().splitlines()[0],
-        'documentation': _utils.format_docstring(s.docstring(raw=True))
-    }
+    try:
+        sig = {
+            'label': s.docstring().splitlines()[0],
+            'documentation': _utils.format_docstring(s.docstring(raw=True))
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @mike-hearn! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #471, opened 2019-01-14
Lazily apply goto_definition settings

gatesn:
`follow_builtin_imports` is only in jedi 0.13.0, we don't want to take this yet until we resolve the issues around spawning too many child procs. 

Until then, we'll make the settings lazy


lgeiger:
> follow_builtin_imports is only in jedi 0.13.0, we don't want to take this yet until we resolve the issues around spawning too many child procs.

If I recall correctly, this was fixed in [`jedi@0.13.2`](https://github.com/davidhalter/jedi/blob/master/CHANGELOG.rst#0132-2018-12-15).

gatesn:
Oh perfect, I'll just bump the min dependency to 0.13.2 then

PR #483, opened 2019-01-25
Turn on 'follow_imports' & 'follow_builtin_imports' options by default (#482)

### LINKED ISSUE #482: [Suggestion] Turn on `follow_imports` and `follow_builtin_imports` options by default
By turning on this options, we can code jump even if you write it like `from xxx import yyy`.
Options included in this PR(#404), but these default values are `null`.
By default, I think that the user can smart code jump.

kashewnuts:
Resolve #482


lgeiger (approved):
LGTM

PR #493, opened 2019-02-10
setup.py: exclude all test packages

Cynerd:
There is a subpackage in test. This subpackage was before this commit
installed although test package was ignored. Extending exclude to ignore
anything that starts with 'test' is enough to exclude any subpackage of
test.

This is causing an error when trying to package pyls in Gentoo. Gentoo
build system disutils eclass checks if there is some test package to be
installed and fails installation step if so. This fixes that.


setup.py
```diff
@@ -25,7 +25,7 @@
 
     # You can just specify the packages manually here if your project is
     # simple. Or you can use find_packages().
-    packages=find_packages(exclude=['contrib', 'docs', 'test']),
+    packages=find_packages(exclude=['contrib', 'docs', 'test*']),
```
gatesn: As mentioned in #495 can we instead use:

```suggestion
    packages=find_packages(exclude=['contrib', 'docs', 'test', 'test.*']),
```

Whichever change is made first wins :)

gatesn (changes requested):

gatesn (approved):

PR #499, opened 2019-02-15
Generate insert snippets for callables

### LINKED ISSUE #496: For completion request return text with argument signature
Hey!

It would be nice if pyls would return the completion type including the argument signature, see the discussion at https://github.com/tigersoldier/company-lsp/issues/29.

Thank you!

gatesn:
Fixes #496

### DIFF
--- a/pyls/lsp.py
+++ b/pyls/lsp.py
@@ -39,6 +39,11 @@ class DiagnosticSeverity(object):
     Hint = 4
 
 
+class InsertTextFormat(object):
+    PlainText = 1
+    Snippet = 2
+
+
 class MessageType(object):
     Error = 1
     Warning = 2
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -6,16 +6,39 @@ log = logging.getLogger(__name__)
 
 
 @hookimpl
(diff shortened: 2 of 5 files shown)


dbarrosop:
I think you should consider disabling this by default and/or documenting in the README how to disable it. This feature is extremely disruptive when coding.

vv111y:
I would like to chime in too. It's a great idea in theory, and I am grateful for your effort and thoughtfulness. It has the potential to become very useful but right now it's really bad.  
Pretty much 100% of my completions I only use a fraction of the arguments. I basically just delete line from cursor, and then I manually enter the rest of the function call.    
It would be good to have a switch to disable, and maybe later to have customization options.

gatesn:
Yup completely agree. Do you think inserting only positional arguments is the right thing, and not optional keyword arguments? Or just disabling entirely?

Shougo:
https://github.com/palantir/python-language-server/pull/526

I think it is the right fix.

vv111y:
Sorry for delay,
I think both: 1) disable option, 2) positional arguments option. 
Another nice option is to have keyword args be available via autocomplete, if you need them, then this cuts down on typing, if not, then just ignore and escape out of the parentheses.

PR #504, opened 2019-02-15
Support yaml config file

### LINKED ISSUE #300: Override pyls configuration from the command line
e.g. `pyls --conf pyls.plugins.pycodestyle.enabled=False`

gatesn:
Initial configuration can be loaded from config file

Fixes #300


youben11:
nice feature! why isn't it merged yet?

PR #505, opened 2019-02-15
Reduce redundant lint calls

lgeiger:
Currently, whenever a file in the workspace changes all open documents are linted. This causes a lot of unnecessary lint calls.
With this PR `pyls` ignores changes to files that do not end with `.py` or `.pyi` and changes that are already handled by the didChange event.

This issue is very problematic when using [`pyls-mypy`](https://github.com/tomv564/pyls-mypy/). `mypy` writes cache files after each lint which results in `didChangeWatchedFiles` to be triggered, resulting in a infinite lint loop.

### DIFF
--- a/pyls/python_ls.py
+++ b/pyls/python_ls.py
@@ -17,6 +17,8 @@ log = logging.getLogger(__name__)
 LINT_DEBOUNCE_S = 0.5  # 500 ms
 PARENT_PROCESS_WATCH_INTERVAL = 10  # 10 s
 MAX_WORKERS = 64
+# We also watch for changes in config files
+PYTHON_FILE_EXTENSIONS = ('.py', '.pyi', 'pycodestyle.cfg', 'setup.cfg', 'tox.ini', '.flake8')
 
 
 class _StreamHandlerWrapper(socketserver.StreamRequestHandler, object):
@@ -295,10 +297,12 @@ class PythonLanguageServer(MethodDispatcher):
             self.lint(doc_uri)
 
     def m_workspace__did_change_watched_files(self, changes=None, **_kwargs):
-        changed_py_files = set(d['uri'] for d in changes if d['uri'].endswith(('.py', '.pyi')))
-        # Only externally changed python files may result in changed diagnostics
+        changed_py_files = set(d['uri'] for d in changes if d['uri'].endswith(PYTHON_FILE_EXTENSIONS))
+        # Only externally changed python files and lint configs may result in changed diagnostics.
         if not changed_py_files:
             return
+        # TODO: We currently don't cache settings therefor we can just lint again.
+        # Here would be the right point to update the settings after a change to config files.
         for doc_uri in self.workspace.documents:
             # Changes in doc_uri are already handled by m_text_document__did_save
             if doc_uri not in changed_py_files:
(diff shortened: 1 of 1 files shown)


lgeiger:
@gatesn Do you have time to review this PR?

pyls/python_ls.py
```diff
@@ -294,10 +294,15 @@ def m_workspace__did_change_configuration(self, settings=None):
         for doc_uri in self.workspace.documents:
             self.lint(doc_uri)
 
-    def m_workspace__did_change_watched_files(self, **_kwargs):
-        # Externally changed files may result in changed diagnostics
+    def m_workspace__did_change_watched_files(self, changes=None, **_kwargs):
+        changed_py_files = set(d['uri'] for d in changes if d['uri'].endswith(('.py', '.pyi')))
```
gatesn: Should be add `.cfg` and `.ini` as well?

pyls/python_ls.py
```diff
@@ -294,10 +294,15 @@ def m_workspace__did_change_configuration(self, settings=None):
...
+        # Only externally changed python files may result in changed diagnostics
+        if not changed_py_files:
+            return
         for doc_uri in self.workspace.documents:
-            self.lint(doc_uri)
+            # Changes in doc_uri are already handled by m_text_document__did_save
+            if doc_uri not in changed_py_files:
```
gatesn: I think this might be the wrong way around?

gatesn (changes requested):
Could you move the list of file extensions to a static at the top? And thoughts about trying to cover some of the linter config files too?

As for the mypy thing, have you tried out https://github.com/palantir/python-language-server/pull/392 ? Wondering whether to invest more time in it

lgeiger:
> As for the mypy thing, have you tried out #392 ? Wondering whether to invest more time in it

I had a few weird errors with https://github.com/tomv564/pyls-mypy since it occasionally writes to stdout, but I didn't had a chance to checkout your PR. I'll try it our soon.

pyls/python_ls.py
lgeiger: I'd expect that this is handled by didChangeConfig, but on the other hand the client doesn't know that we're also accepting config files.

I'll look into what config file changes should trigger a lint.

lgeiger:
> Could you move the list of file extensions to a static at the top? And thoughts about trying to cover some of the linter config files too?

Fixed in 4bfd36acf2dad3979888447b5c7b2d6ab1f3c51c

pyls/python_ls.py
```diff
@@ -294,10 +296,17 @@ def m_workspace__did_change_configuration(self, settings=None):
...
 
-    def m_workspace__did_change_watched_files(self, **_kwargs):
-        # Externally changed files may result in changed diagnostics
+    def m_workspace__did_change_watched_files(self, changes=None, **_kwargs):
+        changed_py_files = set(d['uri'] for d in changes if d['uri'].endswith(PYTHON_FILE_EXTENSIONS))
+        # Only externally changed python files and lint configs may result in changed diagnostics.
+        if not changed_py_files:
+            return
+        # TODO: We currently don't cache settings therefor we can just lint again.
+        # Here would be the right point to update the settings after a change to config files.
```
lgeiger, line 305: I will fix this TODO after this PR has been merged.

gatesn (approved):

PR #509, opened 2019-02-21
Cache settings with lru_cache

### LINKED ISSUE #305: Implement better caching for plugin settings

lgeiger:
`Config.settings()` takes a few milliseconds to run and reads config files from disk. This function runs multiple times during initialization and before every linting request. This means we can very effectively cache the results.

This PR uses `lru_cache` to cache the the result of `Config.settings()`. The cache is invalidated when the config is updated or a config file is changed.

Fixes #305

### DIFF
--- a/pyls/config/config.py
+++ b/pyls/config/config.py
@@ -1,6 +1,10 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import logging
 import pkg_resources
+try:
+    from functools import lru_cache
+except ImportError:
+    from backports.functools_lru_cache import lru_cache
 
 import pluggy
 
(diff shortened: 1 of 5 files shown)


lgeiger:
Hehe, I just noticed that there is already a caching mechanism in place: https://github.com/palantir/python-language-server/blob/develop/pyls/config/source.py#L36
I removed it in the latest commit.

With this PR `config.plugin_settings('pycodestyle')` consistently takes 5-6 µs to execute, while before it took 500 - 1300 µs.

@gatesn What do you think about this approach?

lgeiger:
@gatesn Any thoughts on this?

gatesn (approved):
This looks great, thanks a lot! Sorry for the delay, I've been on vacation for a while

Issue #498, opened 2019-02-14
lint on-save?


dbarrosop:
is it possible to lint on-save only? It's a bit distracting to have messages coming up as you type. Specially as most of them are going to be of the style "invalid syntax" as you are still typing.

PR #511, opened 2019-02-24
allow disabling linting on every change

dbarrosop:
Solves #498 

Not sure if there is a better way of doing but this allows user to configure if they want to lint on every change or not.

### DIFF
--- a/pyls/python_ls.py
+++ b/pyls/python_ls.py
@@ -247,7 +247,9 @@ class PythonLanguageServer(MethodDispatcher):
                 change,
                 version=textDocument.get('version')
             )
-        self.lint(textDocument['uri'])
+        settings = self.config.plugin_settings('jedi_completion')
+        if settings.get('lint_on_change', True):
+            self.lint(textDocument['uri'])
 
     def m_text_document__did_save(self, textDocument=None, **_kwargs):
         self.lint(textDocument['uri'])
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @dbarrosop! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

lgeiger (changes requested):
Thanks for the PR! 
Though I am not sure about this approach. I mainly have two concerns:

1. This PR adds the setting to the `jedi_completion` completion plugin which has nothing to do linting. The alternative would be to add a setting to every linting plugin, which also isn't ideal. 
2. I think this setting belongs to the editor you use and not to the language server itself, since this is a global setting appropriate to every linting provider. E.g. the Atom text editor lets you configure this as via the linter setting.

dbarrosop:
>     1. This PR adds the setting to the `jedi_completion` completion plugin which has nothing to do linting. The alternative would be to add a setting to every linting plugin, which also isn't ideal

I agree, I put it there mostly because most of the code is undocumented and I could find examples on how to do that. Other language servers like `bingo` (golang), let's configure this via a flag; `bingo -diagnostics-style={none|onsave|instant}`. I am personally fine with that approach or with a "global" option. Just need some help/guidance to implement.

>     2. I think this setting belongs to the editor you use and not to the language server itself, since this is a global setting appropriate to every linting provider. E.g. the Atom text editor lets you configure this as via the linter setting.

Not sure I understand what you mean with that. How would the editor set that if not via flag or setting to pyls?

PR #514, opened 2019-03-01
Complete follow_imports defaults change

TalAmuyal:
Hi, first time here :)
Huge thanks for all the hard work, really appreciate it!

As far as I understand, the change in https://github.com/palantir/python-language-server/pull/483 is partial, as it only changes the defaults for the VS-Code plugin, and doesn't change for any other LSP-client.
(I'm using Oni, which is based on Neovim)

I think that this PR completes the change.


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @TalAmuyal! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #525, opened 2019-03-15
Remove unnecessary loops in definition, hover and references

lgeiger:
Just a small cleanup, I noticed while debugging some jedi issues.


ccordoba12 (approved):
Looks good to me, thanks @lgeiger!

PR #538, opened 2019-04-08
Discovery and use of pylint config file for pylint lint plugin

jeeger:
I've added functionality to make the `pylint` plugin find the `pylintrc` config file and use it. I've added a `rcfile` plugin option for the pylint plugin that is populated with the path of any found pylint config file.

Also included is a test that checks whether the config file is passed correctly.

Settings for other plugins are broken out from their config file, but the number of settings for pylint is huge, and passing all options from the config file into the `flags` argument of `Pylint.lint()` would require a huge amount of work. This is a useful first implementation.

### DIFF
--- a/pyls/_utils.py
+++ b/pyls/_utils.py
@@ -70,6 +70,7 @@ def find_parents(root, path, names):
     # Otherwise nothing
     return []
 
+
 def search(root, names, depth=4):
     """Find the first file matching the given names below the given path.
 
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @jeeger! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12:
Closing in favor of #539.

PR #539, opened 2019-04-09
Extend pylint linting integration

ExpectationMax:
Adds the `pyls.plugins.pylint.args` parameter for passing arguments to pylint.
In contrast to #538 this PR does not try to find `.pylintrc` in one of the parent directories and rather suggests to use pass the appropriate path using project specific config files. 

 - Added configuration option `pyls.plugins.pylint.args` to allow passing of arguments to pylint
 - Implemented simple error handling
 - Published configuration options via `vscode-client/package.json`


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @ExpectationMax! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

jeeger:
Ah, this looks interesting as well. I think our two PRs should be integrated. I think that existing config files in the project be reused if possible, that's why I implemented the discovery of `pylintrc` the way I did. 
As my changes only take effect if `configSources` contains "pylint", I think our contributions aren't in conflict and should both be integrated for a better pylint experience!

ExpectationMax:
@jeeger I agree :)
Maybe we should decide on an interface to pass the configuration arguments to pylint in a flexible way.
Then the pylint configuration source can append/prepend the rcfile parameter to the configuration arguments.

Nevertheless, the pylint maintainers have not yet decided on how to handle multiple configuration sources as it seems: https://github.com/PyCQA/pylint/issues/618 thus I would not be sure how we should implement this in particular. If multiple passes of `--rcfile` are supported, then I would suggest the following order: `global_config < manual_config (using configuration interface) < project_config < dir_config`

ccordoba12 (approved):
@ExpectationMax, I think your work is simpler and easier to understand than PR #538.

jeeger:
Should I look into reimplementing config discovery for pylint against this patch, or is this a feature that's not desired/not clear how to implement yet?

ccordoba12:
Sure, but given that PyCQA/pylint#618 is unresolved, I think for now we should only use `~/.pylintrc` or `~/.config/pylintrc` (if that's detected on Linux).

PR #542, opened 2019-04-22
Add `=` to default trigger chars config

goanpeca:
I suggest adding the equal sing to the trigger chars so that the active parameter is correctly updated.

Take for example this code (the * represents the cursor)

```python
def somefunc(param1=0, param2=1, param3=None):
    pass
```

Typing:

```python
somefunc(param2=*
```

Without the `=` sign as trigger, when we are using a parameter that is located somewhere in the call, if we don't add the `=` it won't re-trigger a signature request, and the active parameter will be incorrect. In the example param2 will not be the active one. Adding the = to the list of "trigger" characters solves this.


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @goanpeca! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

goanpeca:
Is there any particular test I should add to this?

ccordoba12:
@gatesn, could take a look at this? We need it for Spyder.

gatesn (approved):

gatesn:
Apologies for the delay on this and other PRs, I will try to look at as many as possible over the next few days.

PR #607, opened 2019-06-21
Don't use Jedi 0.14 for now because it breaks the server

ccordoba12:
I discovered this through the Spyder testing suite this morning. We'll add support for this new version as soon as we can, but for now we need to release a new version of the PyLS that avoids using it.


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @ccordoba12! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #613, opened 2019-07-02
Explicitly define our `TestDocumentSync` capabilities

mpanarin:
According to specification, passing a number in `TextDocumentSync` is for legacy support. I think we should move to an explicit declaration of capabilites, as modern clients require that (ex. Emacs lsp-mode will not send notifications that are not defined in capabilities, which is correct, imo).

### DIFF
--- a/pyls/python_ls.py
+++ b/pyls/python_ls.py
@@ -145,7 +145,13 @@ class PythonLanguageServer(MethodDispatcher):
             'signatureHelpProvider': {
                 'triggerCharacters': ['(', ',', '=']
             },
-            'textDocumentSync': lsp.TextDocumentSyncKind.INCREMENTAL,
+            'textDocumentSync': {
+                'change': lsp.TextDocumentSyncKind.INCREMENTAL,
+                'save': {
+                    'includeText': True,
+                },
+                'openClose': True,
+            },
             'experimental': merge(self._hook('pyls_experimental_capabilities'))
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @mpanarin! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12:

gatesn (approved):

pyls/python_ls.py
```diff
@@ -145,7 +145,12 @@ def capabilities(self):
             'signatureHelpProvider': {
                 'triggerCharacters': ['(', ',', '=']
             },
-            'textDocumentSync': lsp.TextDocumentSyncKind.INCREMENTAL,
+            'textDocumentSync': {
```
gatesn, line 148: I think we want `openClose: true` as well

gatesn (changes requested):

mpanarin:
@gatesn oops, updated.

mpanarin:
@ccordoba12 
ping. As I addressed the comment by @gatesn 
Maybe this can be merged?

ccordoba12:
You're right, thanks for the reminder.

PR #621, opened 2019-07-15
Increase maximum supported version of Jedi

### LINKED ISSUE #606: Completions not working properly after Jedi 0.14.0
Yesterday, python released a new version of jedi 0.14.0. After this autocompletion not working properly. Getting the following error in messages:
`Language Client: Error: AttributeError: 'NoneType' object has no attribute 'full_name'
`

Screen Shot 2019-06-21 at 2 42 08 PM

But working fine on the last release i.e. Jedi 0.13.3

ccordoba12:
Now that Jedi 0.14.1 is out, we can support it.

Fixes #606.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ setup(
         'future>=0.14.0',
         'futures; python_version<"3.2"',
         'backports.functools_lru_cache; python_version<"3.2"',
-        'jedi>=0.13.2,<0.14',
+        'jedi>=0.13.2,<0.15,!=0.14.0',
         'python-jsonrpc-server>=0.1.0',
         'pluggy'
     ],


reneeotten:
I am trying to get this to work in MacPorts where jedi has been updated already to version 0.14.1 (and I would try to avoid downgrading that package again). Only changing the version specifier results in similar test failures as in this PR; is there a clear idea what else would be needed? If so, we could add those patches to MacPorts, until a new version with full support for jedi > 0.14 is released (any time frame for that?). Thanks for all the hard work!

ccordoba12:
The fix is in PR #625 and it'll be available in our next release (0.28)

ccordoba12:
@reneeotten, 0.28.0 is out!

reneeotten:
great, thanks so much @ccordoba12!

PR #623, opened 2019-07-17
Separate signature from docstring on hover

mpanarin:
LSP supports `MarkedString[]` to specify render engine. We can definitely say that function signature should be rendered as python.
For the docstring, as we have no good way of parsing it to guess the engine, we can pass it as markdown.


pyls/plugins/hover.py
```diff
@@ -13,6 +13,15 @@ def pyls_hover(document, position):
     # Find an exact match for a completion
     for d in definitions:
         if d.name == word:
-            return {'contents': _utils.format_docstring(d.docstring()) or ''}
+            doc = _utils.format_docstring(d.docstring()).split('\n\n', 1)
```
ccordoba12: Why is this `split('\n\n', 1)` is needed?

test/plugins/test_hover.py
```diff
@@ -21,7 +21,7 @@ def test_hover():
     doc = Document(DOC_URI, DOC)
 
     assert {
-        'contents': 'main()\n\nhello world'
+        'contents': [{'language': 'python', 'value': 'main()'}, 'hello world']
```
ccordoba12, line 24: This could break existing clients.

pyls/plugins/hover.py
mpanarin: so I can separate signature from docstring

test/plugins/test_hover.py
mpanarin: I highly doubt so. If the client doesn't support it -> it doesn't support lsp specification.
https://microsoft.github.io/language-server-protocol/specification#textDocument_hover

this is a format of `MarkedString[]`

ccordoba12:

andfoy:
I think we should aim to support Markdown fully using ``MarkupContent``, i.e., Finishing https://github.com/palantir/python-language-server/pull/348

test/plugins/test_hover.py
gatesn: I think we have to check the text document client capabilities for `hover.contentFormat`, the protocol certainly didn’t support this at some point in the past

mpanarin: good point, will check into that

mpanarin:
@andfoy Well, that pr implements rst2md converter. Which is wrong, as python docs come in different shape and sizes. And without some specific parsers/converters is there an actual point in that?
While here we basically say render signature as python, everything else as markdown (whatever there is).

gatesn:
I wonder if there's a set of "tell-tale" signs that a docstring is in markdown vs rst vs plain text. In such a way that if those signs aren't there, then it's likely it doesn't matter what format we treat it as.

e.g. `hello` is valid markdown, rst and plain text. Whereas `..<directive>:` is clearly an RST doc.

mpanarin:
@gatesn I updated this pull requrest

In jedi 15.0 author added a proper way to get a signature of a function [here](https://github.com/davidhalter/jedi/blob/master/jedi/api/classes.py#L398) which allows us to grab signature and docstring separately, formatting them accordingly.
Imo, these is reasonable enough change to make the dependency on `jedi >= 0.15.0` as it can definitely be used in `jedi_completion` as well (I didn't check specific use, but I am sure it is there.) making code clearer.
If accepted, supersedes https://github.com/palantir/python-language-server/pull/635

About your comment on `hover.contentFormat` those are the types of `MarkupKind` - plaintext or markdown. So this should work fine as we follow markdown route. 

For a tell-tale signs - this is a good thing for a research. But as I said earlier - it is a huge amount of work, and docstring at the moment are not highlighted properly. Which is a shame.

setup.py
```diff
@@ -36,7 +36,7 @@
         'future>=0.14.0',
         'futures; python_version<"3.2"',
         'backports.functools_lru_cache; python_version<"3.2"',
-        'jedi>=0.13.2,<0.15,!=0.14.0',
+        'jedi>=0.15.0',
```
ccordoba12: Please change this to

    jedi>=0.15.0,<0.16

because Jedi major releases usually break stuff.

test/plugins/test_hover.py
ccordoba12: > this is a format of MarkedString[]

Right, sorry for the noise.

ccordoba12 (commented):
I left a minor comment, otherwise looks good to me.

@gatesn?

setup.py
mpanarin: updated

samiur:
Would love to see this merged, as I'd like the Jedi requirements update

mpanarin:
@ccordoba12 ping on this one )

ccordoba12 (approved):
Thanks for your patience @mpanarin!

max-baz:
Hey guys, would you please cut off a new release targeted the latest python-jedi? It seems [0.29.0 milestone](https://github.com/palantir/python-language-server/milestone/6) is complete anyway 😉

PR #625, opened 2019-07-23
Enable Jedi 0.14.1

### LINKED ISSUE #606: Completions not working properly after Jedi 0.14.0

steff456:
Fix tests and completions for the latest version of jedi.

Fixes #606

### DIFF
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -90,9 +90,9 @@ def _label(definition):
 
 def _detail(definition):
     try:
-        return definition.full_name or ''
+        return definition.parent().full_name or ''
     except AttributeError:
-        return ''
+        return definition.full_name or ''
 
 
 def _sort_text(definition):
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -90,7 +90,7 @@ def _label(definition):
 
 def _detail(definition):
     try:
-        return definition.parent().full_name or ''
+        return definition.full_name or ''
     except AttributeError:
         return ''
 
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -89,7 +89,10 @@ def _label(definition):
 
 
 def _detail(definition):
-    return definition.parent().full_name or ''
+    try:
+        return definition.parent().full_name or ''
+    except AttributeError:
+        return '' 
 
 
 def _sort_text(definition):
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -92,7 +92,7 @@ def _detail(definition):
     try:
         return definition.parent().full_name or ''
     except AttributeError:
-        return '' 
+        return ''
 
 
 def _sort_text(definition):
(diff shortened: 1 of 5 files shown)


pyls/plugins/jedi_completion.py
```diff
@@ -89,7 +89,10 @@ def _label(definition):
 
 
 def _detail(definition):
-    return definition.parent().full_name or ''
+    try:
+        return definition.parent().full_name or ''
+    except AttributeError:
+        return ''
```
ccordoba12: I think this needs to be changed to

    return definition.full_name or ''

for Jedi 0.14, at least according to the tests in

https://github.com/davidhalter/jedi/blob/master/test/test_api/test_full_name.py

test/plugins/test_definitions.py
```diff
@@ -39,7 +39,7 @@ def test_builtin_definition(config):
 
     # No go-to def for builtins
     doc = Document(DOC_URI, DOC)
-    assert [] == pyls_definitions(config, doc, cursor_pos)
+    assert len(pyls_definitions(config, doc, cursor_pos)) == 1
```
ccordoba12, line 42: This fixes the test but fails to catch what the test really is aiming at, i.e. that we shouldn't show definitions for builtin's.

Therefore, please look at why the test is really failing.

test/plugins/test_symbols.py
```diff
@@ -27,7 +27,7 @@ def test_symbols(config):
     symbols = pyls_document_symbols(config, doc)
 
     # All four symbols (import sys, a, B, main, y)
-    assert len(symbols) == 5
+    assert len(symbols) == 4
```
ccordoba12, line 31: Why is the number of symbols different for Jedi 0.14? What symbol is missing now?

ccordoba12 (commented):
@steff456, thanks for finishing this one. I left you some comments for you to take a look at.

test/plugins/test_definitions.py
steff456: This test is failing because jedi is now returning a .pyi file for built-in functions. So now the built-ins are being returned.

test/plugins/test_symbols.py
steff456: The y symbol is not in the root scope, as the option `jedi_symbols: all_scopes` is False it will only return the root symbols. I will let a comment in the test. 
If we want, we can test for all symbols in all the scopes, changing the test or creating a new one.

pyls/plugins/jedi_completion.py
ccordoba12: The change you made was not my idea to fix this. I think to cover both Jedi 0.13 and 0.14, this should be

```python
try:
    return definition.parent().full_name or ''
except AttributeError:
    return definition.full_name or ''
```

ccordoba12 (approved):
Thanks @steff456, great job!

PR #627, opened 2019-07-25
Fix test failures in 0.28.0 with Jedi 0.13

### LINKED ISSUE #626: Test failures in 0.28.0 with Jedi 0.13
Some test failures in the latest release:

```

=================================== FAILURES ===================================
_____________________________ test_jedi_completion _____________________________

config = <pyls.config.config.Config object at 0x7f2aec2207f0>

    def test_jedi_completion(config):
        # Over 'i' in os.path.isabs(...)
        com_position = {'line': 1, 'character': 15}
        doc = Document(DOC_URI, DOC)
        items = pyls_jedi_completions(config, doc, com_position)
    
        assert items
>       assert items[0]['label'] == 'isabs(path)'
E       AssertionError: assert 'isabs(s)' == 'isabs(path)'
E         - isabs(s)
E         + isabs(path)

test/plugins/test_completion.py:52: AssertionError
---------------------------- Captured stderr setup -----------------------------

config = <pyls.config.config.Config object at 0x7f2ae6a98b70>

    def test_builtin_definition(config):
        # Over 'i' in dict
        cursor_pos = {'line': 8, 'character': 24}
    
        # No go-to def for builtins
        doc = Document(DOC_URI, DOC)
>       assert len(pyls_definitions(config, doc, cursor_pos)) == 1
E       AssertionError: assert 0 == 1
E        +  where 0 = len([])
E        +    where [] = pyls_definitions(<pyls.config.config.Config object at 0x7f2ae6a98b70>, <pyls.workspace.Document object at 0x7f2ae6a86940>, {'character': 24, 'line': 8})

test/plugins/test_definitions.py:42: AssertionError



config = <pyls.config.config.Config object at 0x7f2ae68d5400>

    def test_symbols(config):
        doc = Document(DOC_URI, DOC)
        config.update({'plugins': {'jedi_symbols': {'all_scopes': False}}})
        symbols = pyls_document_symbols(config, doc)
    
        # All four symbols (import sys, a, B, main)
        # y is not in the root scope, it shouldn't be returned
>       assert len(symbols) == 4
E       AssertionError: assert 5 == 4
E        +  where 5 = len([{'containerName': None, 'kind': 2, 'location': {'range': {'end': {'character': 10, 'line': 0}, 'start': {'character':...: 'file:///build/python-language-server/src/python-language-server-0.28.0/test/plugins/test_symbols.py'}, 'name': 'y'}])

test/plugins/test_symbols.py:31: AssertionError
```

steff456:
Fixes #626

### DIFF
--- a/test/plugins/test_completion.py
+++ b/test/plugins/test_completion.py
@@ -1,10 +1,13 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import os
+import jedi
+import pytest
 
 from pyls import uris, lsp
 from pyls.workspace import Document
 from pyls.plugins.jedi_completion import pyls_completions as pyls_jedi_completions
 from pyls.plugins.rope_completion import pyls_completions as pyls_rope_completions
+from test.test_utils import version_str2tuple as str2tuple
 
 LOCATION = os.path.realpath(
     os.path.join(os.getcwd(), os.path.dirname(__file__))
@@ -42,6 +45,8 @@ def test_rope_import_completion(config, workspace):
     assert items is None
 
 
+@pytest.mark.skipif(str2tuple(jedi.__version__) <= str2tuple('14.0'),
+                    reason='This test fails with previous versions of jedi')
 def test_jedi_completion(config):
     # Over 'i' in os.path.isabs(...)
     com_position = {'line': 1, 'character': 15}
--- a/test/plugins/test_completion.py
+++ b/test/plugins/test_completion.py
@@ -1,5 +1,6 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import os
+from test.test_utils import version_str2tuple as str2tuple
 import jedi
 import pytest
 
@@ -7,7 +8,6 @@ from pyls import uris, lsp
 from pyls.workspace import Document
 from pyls.plugins.jedi_completion import pyls_completions as pyls_jedi_completions
 from pyls.plugins.rope_completion import pyls_completions as pyls_rope_completions
-from test.test_utils import version_str2tuple as str2tuple
 
 LOCATION = os.path.realpath(
     os.path.join(os.getcwd(), os.path.dirname(__file__))
--- a/test/plugins/test_completion.py
+++ b/test/plugins/test_completion.py
@@ -1,8 +1,8 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import os
-from test.test_utils import version_str2tuple as str2tuple
 import jedi
 import pytest
+from distutils.version import LooseVersion
 
 from pyls import uris, lsp
 from pyls.workspace import Document
@@ -45,7 +45,7 @@ def test_rope_import_completion(config, workspace):
     assert items is None
 
 
-@pytest.mark.skipif(str2tuple(jedi.__version__) <= str2tuple('14.0'),
+@pytest.mark.skipif(LooseVersion(jedi.__version__) < LooseVersion('14.0'),
                     reason='This test fails with previous versions of jedi')
 def test_jedi_completion(config):
     # Over 'i' in os.path.isabs(...)
--- a/test/plugins/test_completion.py
+++ b/test/plugins/test_completion.py
@@ -45,7 +45,7 @@ def test_rope_import_completion(config, workspace):
     assert items is None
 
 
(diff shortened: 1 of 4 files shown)


test/plugins/test_completion.py
```diff
@@ -42,6 +45,8 @@ def test_rope_import_completion(config, workspace):
     assert items is None
 
 
+@pytest.mark.skipif(str2tuple(jedi.__version__) <= str2tuple('14.0'),
```
ccordoba12: You can compare strings directly, like this

    jedi.__version__ < '0.14.0'

so there's no need of `str2tuple`.

Also, please notice the `<` instead of `<=` here.

test/plugins/test_definitions.py
```diff
@@ -33,6 +38,8 @@ def test_definitions(config):
     assert [{'uri': DOC_URI, 'range': def_range}] == pyls_definitions(config, doc, cursor_pos)
 
 
+@pytest.mark.skipif(str2tuple(jedi.__version__) <= str2tuple('14.0'),
```
ccordoba12: Please compare strings directly here.

test/plugins/test_symbols.py
```diff
@@ -21,6 +25,8 @@ def main(x):
 """
 
 
+@pytest.mark.skipif(str2tuple(jedi.__version__) <= str2tuple('14.0'),
```
ccordoba12: Please compare strings directly here.

test/test_utils.py
```diff
@@ -87,3 +87,8 @@ def test_clip_column():
     assert _utils.clip_column(2, ['123\n', '123'], 0) == 2
     assert _utils.clip_column(3, ['123\n', '123'], 0) == 3
     assert _utils.clip_column(4, ['123\n', '123'], 1) == 3
+
+
+def version_str2tuple(version_str):
```
ccordoba12: Please remove this function.

ccordoba12 (commented):
Thanks for this @steff456. I left you some comments to simplify the Jedi version comparison.

test/plugins/test_completion.py
steff456: We cannot compare the strings as that. For example,

Captura de pantalla 2019-07-25 a la(s) 2 23 39 p  m

I thought that we shouldn't run those tests with jedi 0.14.0, only with greater versions.

steff456: We cannot compare version strings as that, in some cases it will not work as expected
Captura de pantalla 2019-07-25 a la(s) 2 23 39 p  m

I thought we only wanted to run these tests if jedi version is 0.14.1 or greater

ccordoba12: You're totally right! Then please use the following function from distutils, which should be more robust than str2tuple:

```python
from distutils.version import LooseVersion

>>> LooseVersion('0.3.0') < LooseVersion('0.14.0')
True
```

test/plugins/test_completion.py
```diff
@@ -42,6 +45,8 @@ def test_rope_import_completion(config, workspace):
     assert items is None
 
 
+@pytest.mark.skipif(LooseVersion(jedi.__version__) < LooseVersion('14.0'),
```
ccordoba12: Instead of `14.0` here, you need to use `0.14.0`.

test/plugins/test_definitions.py
```diff
@@ -33,6 +38,8 @@ def test_definitions(config):
     assert [{'uri': DOC_URI, 'range': def_range}] == pyls_definitions(config, doc, cursor_pos)
 
 
+@pytest.mark.skipif(LooseVersion(jedi.__version__) < LooseVersion('14.0'),
```
ccordoba12: Also here `0.14.0`.

test/plugins/test_symbols.py
```diff
@@ -21,6 +25,8 @@ def main(x):
 """
 
 
+@pytest.mark.skipif(LooseVersion(jedi.__version__) < LooseVersion('14.0'),
```
ccordoba12: And here `0.14.0`.

ccordoba12 (commented):
Thanks @steff456! Last comments and then this should be ready.

ccordoba12 (approved):
Thanks @steff456!

PR #630, opened 2019-08-04
Adding Docker image for running the pyls server

youben11:
# Docker image for pyls
This will enable running pyls (tcp mode) for dev, testing and prod in the same environment.

This currently uses the pyls version available on pip, but we can extend it to have the nightly version for testing and developing on the last version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM python:3.6-slim
+# Install pyls
+RUN pip install python-language-server
+# Run the LSP on port 5007
+ENTRYPOINT ["pyls", "--tcp", "--port", "5007", "--host", "0.0.0.0"]


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @youben11! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12:
I really don't think this belongs here, sorry. We don't use Docker at all because things are developed in the same machine (even for tcp).

@gatesn, thoughts about this?

youben11:
Thanks for answering @ccordoba12 , I personally like when projects have a Dockerfile so I can run, test and modify them without the fear of breaking something or other dependencies, and I run containers for every application I use (on my local machine). And the way I got to this PR is that I wanted to start working on pyls and needed that environment, so I thought of sharing it.

ccordoba12:
Ok, I understand. But you can also use virtualenvs or venvs, which are simpler to setup up and work in all platforms.

Since this is mostly a convenience for your development workflow, I'm closing it, sorry.

PR #642, opened 2019-08-30
Enable parent process checking for TCP server

andfoy:
Allow to shutdown pyls if TCP mode is active and the parent process reported by the ```$/initialized``` call is killed.

### DIFF
--- a/pyls/python_ls.py
+++ b/pyls/python_ls.py
@@ -2,6 +2,7 @@
 import logging
 import socketserver
 import threading
+from functools import partial
 
 from pyls_jsonrpc.dispatchers import MethodDispatcher
 from pyls_jsonrpc.endpoint import Endpoint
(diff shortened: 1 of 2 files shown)


pyls/python_ls.py
```diff
@@ -35,15 +36,17 @@ def handle(self):
         self.delegate.start()
 
 
-def start_tcp_lang_server(bind_addr, port, handler_class):
+def start_tcp_lang_server(bind_addr, port, check_parent_process,
+                          handler_class):
```
ccordoba12: Let's leave this in a single line:

    def start_tcp_lang_server(bind_addr, port, check_parent_process, handler_class):

ccordoba12 (commented):
I left a minor comment, otherwise looks good.

PR #654, opened 2019-09-11
Don't limit jedi to <0.15

mcepl:
Tests pass without failure (82 passed, 5 skipped), and it improves my life (a Linux Python maintainer) so I use system jedi in openSUSE. I will create the system package of pyls.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ setup(
         'future>=0.14.0',
         'futures; python_version<"3.2"',
         'backports.functools_lru_cache; python_version<"3.2"',
-        'jedi>=0.14.1,<0.15',
+        'jedi>=0.14.1',
         'python-jsonrpc-server>=0.1.0',
         'pluggy'
     ],


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @mcepl! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12:
The update to Jedi 0.15 is being handled in PR #623, which we haven't had time to test with Spyder (I'm the Spyder maintainer and started to co-maintain this project as part of our move to use the Language Server Protocol in our Editor).

gatesn:
@ccordoba12 what do you think of just not placing an upper bound on Jedi and fixing when issues arise? The API surface we use shouldn’t change too frequently. As far as I’m aware the breaks we’ve seen have been with the new environment APIs which seem to have somewhat stabilised now.

I’ll allow @davidhalter to comment, but I don’t believe it’s his intention to break with _every_ minor version - even if strictly speaking it is allowed by semver pre-1.0.

Perhaps it is time to push for a Jedi 1.0?

ccordoba12:
@gatesn, we've been using Jedi for a long time (since version 0.8, I think) and usually major releases (e.g. from 0.10 to 0.11) break stuff.

Since we provide an automatic way for Spyder users to report errors to our Github issue tracker, we've been flooded by Jedi issues (say 20 to 30 issues per day for three or four days) when a new version comes up. That's why I'd like to be on the conservative side here.

However, in order to not steer this project only towards Spyder needs, I'll add a requirement in Spyder to not use a new PyLS version until we properly test it, I'll merge #623 (which agrees with the LSP reference just fine) and release 0.29 with it.

davidhalter:
I totally understand not updating Jedi automatically. IMO it hasn't been an issue since the environemt changes (0.11 to 0.12 I think), but there might still be issues if you use a different feature set than other solutions.

So I would probably not auto-upgrade, but it's generally a good idea to upgrade as soon as possible, because the upgrades usually contain bug fixes and new cool features. :)

PR #655, opened 2019-09-12
Shutdown TCP ServerSocket when parent process ends

andfoy:
This PR enables to interrupt actual TCP server when parent server ends, as the current behaviour only closes the communication channel

### DIFF
--- a/pyls/_utils.py
+++ b/pyls/_utils.py
@@ -152,7 +152,7 @@ def interrupt_process(pid=None):
         pid = os.getpid()
     sig = signal.SIGINT
     if os.name == 'nt':
-        sig = signal.CTRL_C_EVENT
+        sig = signal.CTRL_C_EVENT  # pylint: disable=no-member
     os.kill(pid, sig)
 
 
--- a/pyls/_utils.py
+++ b/pyls/_utils.py
@@ -4,7 +4,6 @@ import inspect
 import logging
 import os
 import sys
-import signal
 import threading
 
 PY2 = sys.version_info.major == 2
@@ -145,15 +144,6 @@ def clip_column(column, lines, line_number):
     # https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#position
     max_column = len(lines[line_number].rstrip('\r\n')) if len(lines) > line_number else 0
     return min(column, max_column)
-
-
-def interrupt_process(pid=None):
-    if pid is None:
-        pid = os.getpid()
-    sig = signal.SIGINT
-    if os.name == 'nt':
-        sig = signal.CTRL_C_EVENT  # pylint: disable=no-member
-    os.kill(pid, sig)
 
 
 if os.name == 'nt':
--- a/pyls/_utils.py
+++ b/pyls/_utils.py
@@ -4,6 +4,7 @@ import inspect
 import logging
 import os
 import sys
+import signal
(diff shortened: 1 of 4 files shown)


pyls/_utils.py
```diff
@@ -146,6 +147,15 @@ def clip_column(column, lines, line_number):
     return min(column, max_column)
 
 
+def interrupt_process(pid=None):
```
gatesn: Why are we sending a sigint to ourselves? Python’s TCP server polls for shutdown by default every 500ms - can’t we just call shutdown on the server?

https://docs.python.org/2/library/socketserver.html#SocketServer.BaseServer.serve_forever

pyls/python_ls.py
```diff
@@ -120,6 +123,9 @@ def m_exit(self, **_kwargs):
         self._jsonrpc_stream_reader.close()
         self._jsonrpc_stream_writer.close()
 
+        if self._is_tcp and self._check_parent_process:
```
gatesn: I don’t like forking the logic here. If there’s a need to send a sigint to ourselves then why not do it for both stdio and tcp?

pyls/_utils.py
andfoy: For some reason, the interrupt is not working, I'll take a look more carefully, I'm aware this is not the most appropriate way of handling this

pyls/python_ls.py
andfoy: See above.

ccordoba12:
@gatesn, I think the last commit by @andfoy implements a solution that meets your requirements. Do you agree with it?

pyls/python_ls.py
```diff
@@ -187,19 +198,19 @@ def m_initialize(self, processId=None, rootUri=None, rootPath=None, initializati
...
                     log.info("parent process %s is not alive", pid)
                     self.m_exit()
-                log.debug("parent process %s is still alive", pid)
-                threading.Timer(PARENT_PROCESS_WATCH_INTERVAL, watch_parent_process, args=[pid]).start()
-
-            watching_thread = threading.Thread(target=watch_parent_process, args=(processId,))
-            watching_thread.daemon = True
-            watching_thread.start()
+                else:
+                    log.debug("parent process %s is still alive", pid)
```
ccordoba12: Please remove this `log` entry to not fill the PyLS log with useless info.

ccordoba12 (commented):
@andfoy, i left you a small comment, then this should be ready.

ccordoba12:
@gatesn, we really need this in Spyder, so if I don't hear from you until Tuesday I'll merge it. Hope that's ok for you.

test/test_language_server.py
```diff
@@ -72,12 +76,17 @@ def test_initialize(client_server):  # pylint: disable=redefined-outer-name
 @unix_only
```
ccordoba12, line 76: @andfoy, could we remove this decorator to see if things as expected on Windows? I mean, I think after PR #445 they should.

andfoy: The problem is that tests fail when we try to use multiprocessing on Windows

ccordoba12: ok, no problem then :+1:

ccordoba12 (approved):
Thanks @andfoy!

Issue #190, opened 2017-11-14
Support Flake8 plugins and configuration


sivakov512:
Can I use flake8 tool for linting? I looked in [package.json](https://github.com/palantir/python-language-server/blob/develop/vscode-client/package.json#L22), but did not find there the corresponding options.

if I change `configurationSources` to ['flake'], then I do not see any changes and `pyflakes` continues to be used. I store `flake8` settings in .`flake8` file for all projects.

If this is important, I use [lsp-mode](https://github.com/emacs-lsp/lsp-mode) in Emacs and [LanguageClient-neovim](https://github.com/autozimu/LanguageClient-neovim) in Neovim as clients.

gatesn:
So you need to set `configurationSources` to `['flake8']`. Then we will parse your flake8 config and pass the equivalent settings into pycodestyle and pyflakes. We don't use flake8 directly.

sivakov512:
I try it and nothing changed. This option parse `.flake8` file?

sivakov512:
I have a `.flake8` file with this content
```
[flake8]
ignore = E501, E265, F811
max-line-length = 160
```
But I still get errors about the length of the line of more than 79 characters. A similar problem if the file is named `flake8.cfg`.

Why not use flake8 directly (as optional feature, for example)? I use some flake8-plugins, for example, `flake8-isort` and `flake8-pdb` and `pyls` do not work with them

gatesn:
The reasoning is that there are _lots_ of tools for python linting, and to avoid keeping up with all of them, we would just integrate with the core set of tools that actually do the lifting. These being pycodestyle, pydocstyle, pyflakes, pylint etc.

Aggregation tools, such as flake8, pylama etc. tend to add lots of duplication.

However, you raise a very good point around plugins for these tools, as well as the complexity we now have to deal with around configuration options. I'll give it a think, but it may just be easier to provide plugins for these tools and then allow the user to enable/disable plugins to get their preferred set of tools.

purpleP:
@gatesn If I'm installed python-language-server via pypi where do I change `configurationSources`?

gatesn:
Depends which client you’re using?

________________________________

purpleP:
@gatesn neovim language server.

gatesn:
I think you'll have to take a look at this: https://github.com/autozimu/LanguageClient-neovim/issues/139

purpleP:
@gatesn 
[...]

PR #656, opened 2019-09-19
Flake8 plugin for linting

youben11:
As per the discussion here #190 a way of supporting flake8 linting is by implementing an additional plugin that will use the flake8 config directly instead of passing it to other linting tools like pycodestyle and pyflakes (even if flake8 is a wrapper around those tools by itself).

Flake8 lacks (for now) of a stable [public python API](http://flake8.pycqa.org/en/latest/user/python-api.html) which should be implemented in the future, so I figured out a dirty way from the source code to access the report results and get the information we need to report back to the IDE, however, this should be updated whenever a python API is made available.

### DIFF
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -1,0 +1,61 @@
+# Copyright 2017 Palantir Technologies, Inc.
+import logging
+from flake8.api import legacy as flake8
+from pyls import hookimpl, lsp
+
+log = logging.getLogger(__name__)
+
+
+@hookimpl
+def pyls_lint(config, document):
+    settings = config.plugin_settings('flake8')
+    log.debug("Got flake8 settings: %s", settings)
+
+    opts = {
+        'exclude': settings.get('exclude'),
+        'filename': settings.get('filename'),
+        'hang_closing': settings.get('hangClosing'),
+        'ignore': settings.get('ignore'),
+        'max_line_length': settings.get('maxLineLength'),
+        'select': settings.get('select'),
+    }
+
+    kwargs = {k: v for k, v in opts.items() if v}
+    style_guide = flake8.get_style_guide(quiet=4, verbose=0, **kwargs)
+
+    report = style_guide.check_files([document.path])
+    diagnostics = parse_report(document, report)
+
+    return diagnostics
+
+
+def parse_report(document, report):
+    file_checkers = report._application.file_checker_manager.checkers
+    # There should be only one filechecker since we are parsing using a path and not a pattern
+    if len(file_checkers) > 1:
+        log.error("Flake8 parsed more than a file for '%s'", document.path)
+
+    diagnostics = []
+    file_checker = file_checkers[0]
+    for error in file_checker.results:
+        code, line, character, msg, physical_line = error
+        diagnostics.append(
+            {
+                'source': 'flake8',
+                'code': code,
+                'range': {
+                    'start': {
+                        'line': line,
+                        'character': character
+                    },
+                    'end': {
+                        'line': line,
+                        'character': len(physical_line)
+                    }
+                },
+                'message': msg,
+                'severity': lsp.DiagnosticSeverity.Warning,
+            }
+        )
+
+    return diagnostics
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Palantir Technologies, Inc.
+# Copyright 2019 Palantir Technologies, Inc.
 import logging
 from flake8.api import legacy as flake8
 from pyls import hookimpl, lsp
@@ -31,7 +31,7 @@ def pyls_lint(config, document):
 
 def parse_report(document, report):
     file_checkers = report._application.file_checker_manager.checkers
-    # There should be only one filechecker since we are parsing using a path and not a pattern
+    # There should be only a filechecker since we are parsing using a path and not a pattern
     if len(file_checkers) > 1:
         log.error("Flake8 parsed more than a file for '%s'", document.path)
 
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -21,6 +21,7 @@ def pyls_lint(config, document):
         'select': settings.get('select'),
     }
 
+    # Build the flake8 checker and use it to generate a report from the document
     kwargs = {k: v for k, v in opts.items() if v}
     style_guide = flake8.get_style_guide(quiet=4, verbose=0, **kwargs)
     report = style_guide.check_files([document.path])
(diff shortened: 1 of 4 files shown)


youben11:
@gatesn

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2017 Palantir Technologies, Inc.
```
uschi2000: 2019

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,61 @@
...
+import logging
+from flake8.api import legacy as flake8
+from pyls import hookimpl, lsp
+
+log = logging.getLogger(__name__)
+
+
```
uschi2000, line 9: can you add some dev docs, please: what is this, what does it do? please also think about user-facing documentation (@gatesn , where do we keep those?)

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,61 @@
...
+        'select': settings.get('select'),
+    }
+
+    kwargs = {k: v for k, v in opts.items() if v}
+    style_guide = flake8.get_style_guide(quiet=4, verbose=0, **kwargs)
+
+    report = style_guide.check_files([document.path])
+    diagnostics = parse_report(document, report)
+
+    return diagnostics
```
uschi2000: inline return statement

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,61 @@
...
+
+@hookimpl
+def pyls_lint(config, document):
+    settings = config.plugin_settings('flake8')
+    log.debug("Got flake8 settings: %s", settings)
+
+    opts = {
+        'exclude': settings.get('exclude'),
+        'filename': settings.get('filename'),
+        'hang_closing': settings.get('hangClosing'),
```
uschi2000, line 24: what's the deal with camelCase vs snake_case here?

setup.py
```diff
@@ -76,6 +76,7 @@
         ],
         'pyls': [
             'autopep8 = pyls.plugins.autopep8_format',
+            'flake8 = pyls.plugins.flake8_lint',
```
uschi2000, line 81: possible to add a test for this functionality?

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def parse_report(document, report):
+    file_checkers = report._application.file_checker_manager.checkers
+    # There should be only one filechecker since we are parsing using a path and not a pattern
+    if len(file_checkers) > 1:
+        log.error("Flake8 parsed more than a file for '%s'", document.path)
+
+    diagnostics = []
+    file_checker = file_checkers[0]
```
uschi2000, line 75: do you need to guard against `len(file_checkers) == 0`, or are you guaranteed that there is one element at least?

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,61 @@
...
+    diagnostics = parse_report(document, report)
+
+    return diagnostics
+
+
+def parse_report(document, report):
+    file_checkers = report._application.file_checker_manager.checkers
+    # There should be only one filechecker since we are parsing using a path and not a pattern
+    if len(file_checkers) > 1:
+        log.error("Flake8 parsed more than a file for '%s'", document.path)
```
uschi2000, line 72: s/a file/one file/

pyls/plugins/flake8_lint.py
youben11: filecheckers is a list of checkers for files that are being linted, since the pyls_lint hook should receive only one file to lint, having more than one is incorrect, I should also check for `len(file_checkers) == 0` to return an empty list also

pyls/plugins/flake8_lint.py
youben11: I got that part from another plugin using the same conf, but it looks like it wasn't working, so I'm willing to debug this to fix it

pyls/plugins/flake8_lint.py
youben11: of course I will

pyls/plugins/flake8_lint.py
gatesn: VSCode settings convention appears to be camel case, but we need to pass it to the Flake8 as a snake case option

youben11: @gatesn how does it affects VSCode if for example we rename plugins.pycodestyle.hangClosing to plugins.pycodestyle.hang-closing ?

youben11:
Test is failing because of the latest release of pylint  #661

setup.py
youben11: done

pyls/plugins/flake8_lint.py
gatesn: Renaming configs would likely break existing users. We could build utility functions to convert between case styles, but not sure it's that bad as-is

uschi2000:
Looks good to me!

pyls/plugins/flake8_lint.py
youben11: got it

youben11:
@gatesn As flake8_config is now aggregating flake8 config for both pycodestyle and mccabe, we should update it to load the config under plugins.falke8, right? Users will no longer be able to use flake8 config files with pycodestyle and mccabe that way, the flake8 plugin will be the only one to use flake8 config files.

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,93 @@
...
+        report: A Report object returned by checking the document.
+    Returns:
+        A list of dictionaries.
+    """
+
+    file_checkers = report._application.file_checker_manager.checkers
+    # No file have been checked
+    if not file_checkers:
+        return []
+    # There should be only a filechecker since we are parsing using a path and not a pattern
```
gatesn, line 70: We can collapse these two statements into `if len(blah) != 1`

setup.py
```diff
@@ -48,6 +48,7 @@
     extras_require={
         'all': [
             'autopep8',
+            'flake8==3.7.8',
```
gatesn: Even though we use internal APIs, let’s not pin this. Otherwise we’ll end up conflicting with people’s own Flake8 versions.

We should just track latest the best we can I reckon

pyls/plugins/flake8_lint.py
```diff
@@ -0,0 +1,93 @@
...
+"""Linter pluging for flake8"""
+import logging
+from flake8.api import legacy as flake8
+from pyls import hookimpl, lsp
+
+log = logging.getLogger(__name__)
+
+
+@hookimpl
+def pyls_lint(config, document):
```
gatesn, line 17: Should we default this plugin to disabled, and when it is enabled disable pep8 and pyflakes? I think at the moment we’d end up with duplicate diagnostics which could be annoying

pyls/plugins/flake8_lint.py
youben11: I don't see the way we can do that because they are totally different. The first reflects the case where a file being linted haven't been saved yet so we should return no errors, the second might happen if the filename was interpreted as a pattern, I did this check cause the internal API may receive both filename and pattern so we must log the case where it's not working as expected.

pyls/plugins/flake8_lint.py
youben11: yeah, they shouldn't be running at the same time, and disabling flake8 by default is a good idea so current users don't feel the change but can enable linting directly via flake8 if they want to. But flake8_config have to be made loading config into flake8 plugin as well from now on.

setup.py
youben11: done

gatesn (approved):

PR #657, opened 2019-09-22
Support pyenv virtualenv

15cm:
Add pyenv support so that things like ".python-version" works. The first path in Document._extra_sys_path will be used to find the virtualenv via `pyenv prefix $path`.

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -222,11 +222,9 @@ class Document(object):
         sys_paths = set()
         # TODO(gatesn): #339 - make better use of jedi environments, they seem pretty powerful
         # Support pyenv virtualenv for LSP project roots
-        if distutils.spawn.find_executable(
-                'pyenv') and self._extra_sys_path and len(
-                    self._extra_sys_path) > 0:
+        if distutils.spawn.find_executable('pyenv') and self._extra_sys_path:
 
-            class enter_pyenv_shell_env():
+            class enter_pyenv_shell_env(object):
                 def __init__(self):
                     self.env_pyenv_keys = ['PYENV_VERSION', 'PYENV_DIR']
                     self.env_pyenv_old_vals = []
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -1,12 +1,14 @@
 # Copyright 2017 Palantir Technologies, Inc.
+import distutils.spawn
 import io
 import logging
 import os
 import re
+import subprocess
 
 import jedi
 
-from . import lsp, uris, _utils
+from . import _utils, lsp, uris
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @15cm! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

gatesn:
I don't think I understand your use-case. What client are you using? Why would you like to add site-packages of a different Python environment to the language server vs running the language server _in_ that environment?

15cm:
I am using lsp-mode on Emacs. If I follow the latter way, I need to have
pyls installed in every environment. That's acceptable to me but I also
need to know which pyls to launch in lsp-mode. That's the issue for me. My
goal is that we have one pyls server and documents will adjust the env
according to the pyenv they are in.

15cm:
Made a mistake in the goal part. Let me repharse my goal: have one pyls
entry and have corresponding environment in each pyls process that's
launched by different projects.

ccordoba12:
> have one pyls entry and have corresponding environment in each pyls process that's launched by different projects.

Then this is not the right approach. You need to add support for [Jedi environments](https://jedi.readthedocs.io/en/latest/_modules/jedi/api/environment.html) here instead.

15cm:
For _pyls_, [Jedi Environments](https://jedi.readthedocs.io/en/latest/_modules/jedi/api/environment.html) only provides some utility function and class, I think anyway we should modify the `sys_path` that's passed to Jedi to support _pyenv_ since only _pyls_ have the project root for a `document`.
How do you think if I add pyenv support in _Jedi_ and we do something like this in _pyls_?
```python
def sys_path(self):
    # Some check...
    project_root = self._extra_sys_path[0]
    path = list(self._extra_sys_path)
    environment = jedi.api.environment.get_cached_environment_for_project(project_root)
    path.extend(environment.get_sys_path())
    return path
```

bersace:
I confirm that I expect pyls to respect `.python-version` while analyzing a project.

@ccordoba12 how do you manage venv/python version in different project and pyls ? Say having python2 and python3 projects ?

bersace:
See here http://www.lexev.org/en/2019/sublime-text-and-language-server-protocol-lsp/ . One sets `VIRTUAL_ENV` envvar in Sublime . Maybe is sublime running a pyls process per project ?

ccordoba12:
@bersace, see PR #680.

15cm:
As PR #680 is on the same path as I originally proposed , any plan to support `pyenv` in one `pyls` process? As the `pyenv` part have been already done in this PR, I don't mind porting it to the new code.

bersace:
@ccordoba12 Thanks ! If i understand correctly : lsp-mode must configure pyls with extra path from pyenv ?

ccordoba12:
Not the extra path of the intepreter's `sys.path`, but the path to interpreter directly in the `jedi.environment` variable:

https://github.com/palantir/python-language-server/pull/680/files#diff-942a8b6d9b93cba7f0d40bd9db4a4d2fR37

syphar:
@ccordoba12 when debugging my problem I stumbled on this PR, 

for me pyls doesn't work when using a pyenv-virtualenv, but is fine when using a virtualenv on the same python version/executable than the global pyls. 

How to set/use `jedi.api.environemnt`? 

(I'm using `LanguageClient-neovim`)

ccordoba12:
You need to set the Python interpreter of your env in the `jedi.environment` option, then send a `workspace/didChangeConfiguration` to the server.

However, I don't know how that's done in `LanguageClient-neovim`, sorry.

syphar:
Thank you very much @ccordoba12 ! I'll figure it out.

15cm:
Thanks for your work @ccordoba12! It's a nice implementation that delegates the environment passing to LSP clients.

ccordoba12:
Actually that was done by @goanpeca. And we're already using it successfully in Spyder.

carter54:
@15cm @ccordoba12 @syphar 
Hi, thanks for this work. I am wondering that can we pass the current using python interpreter to Jedi environment throught workspace/didChangeConfiguration or we need to set jedi environment manually and use workspace/didChangeConfiguration to tell lsp server the config has been changed? Thx!

carter54:
> @15cm @ccordoba12 @syphar
> Hi, thanks for this work. I am wondering that can we pass the current using python interpreter to Jedi environment throught workspace/didChangeConfiguration or we need to set jedi environment manually and use workspace/didChangeConfiguration to tell lsp server the config has been changed? Thx!

I have figured out this, workspace/didChangeConfiguration can send Jedi environment to the server, and server will change the python interpreter in jedi.environment! Great job! Thx!

PR #663, opened 2019-09-30
Fix failing pylint test

### LINKED ISSUE #661: tests failing because of latest pylint release
Looks like the latest release of pylint is more precise with the column where the syntax error is.

### The test

We are testing pylint on the following document. (full code [here](https://github.com/palantir/python-language-server/blob/c3cab77a85b1de4af1aec1bafea6a7320d6baec5/test/plugins/test_pylint_lint.py))

```python
def hello()
    pass
```

pylint 2.3.1 reports back the 0 as the character where the syntax error is, however, pylint 2.4.1 is reporting back a 12 which is actually the exact location where a ':' must be introduced. 

The test is still expecting a value of 0 so the test fails.

youben11:
fixing failing test because of the latest release of pylint #661 

Fixes #661.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,6 +1,10 @@
 # Copyright 2017 Palantir Technologies, Inc.
+import sys
 import pytest
 from pyls import IS_WIN
 
+IS_PY3 = sys.version_info.major == 3
+
 unix_only = pytest.mark.skipif(IS_WIN, reason="Unix only")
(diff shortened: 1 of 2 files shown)


youben11:
Looks like pylint don't support python2.7, so the latest release only affected python3 and the python2 test now fails because they have different behaviors (each version return a different result)

ccordoba12 (approved):
Looks good to me, thanks @youben11!

PR #665, opened 2019-09-30
Handle textDocument/foldingRange call

andfoy:
This PR enables pyls to handle and compute folding ranges for a given document

### DIFF
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -68,6 +68,11 @@ def pyls_experimental_capabilities(config, workspace):
 
 
 @hookspec(firstresult=True)
+def pyls_folding_range(config, workspace, document):
+    pass
+
+
+@hookspec(firstresult=True)
 def pyls_format_document(config, workspace, document):
     pass
 
@@ -109,9 +114,4 @@ def pyls_settings(config):
 
 @hookspec(firstresult=True)
 def pyls_signature_help(config, workspace, document, position):
-    pass
-
-
-@hookspec(firstresult=True)
-def pyls_folding_range(config, workspace, document):
     pass
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -110,3 +110,8 @@ def pyls_settings(config):
 @hookspec(firstresult=True)
 def pyls_signature_help(config, workspace, document, position):
     pass
+
+
+@hookspec(firstresult=True)
+def pyls_folding_range(config, workspace, document):
+    pass
--- a/pyls/plugins/folding.py
+++ b/pyls/plugins/folding.py
@@ -1,0 +1,76 @@
+
+import ast
+from pyls import hookimpl
+
+
+@hookimpl
+def pyls_folding_range(document):
+    program = str(document.source)
+    # Add an additional "pass" to ensure that there's always a next node
+    program = program + '\n\npass'
+    tree = None
+    lines = []
+    while tree is None:
+        try:
+            tree = ast.parse(program)
+        except SyntaxError as e:
+            offending_line = e.lineno
+            lines = program.splitlines(True)
+            lines = lines[:offending_line - 1]
+            program = ''.join(lines)
+
+    folding_ranges = []
+    folding_starts = {}
+
+    stack = [(tree, [])]
+    while len(stack) > 0:
+        node, ctx = stack.pop(0)
+        node_id = id(node)
+        if hasattr(node, 'lineno'):
+            line, col = node.lineno, node.col_offset
+            folding_starts[node_id] = (line, col)
+
+        nodes = []
+        new_ctx = [node_id] + ctx
+        for child in ast.iter_child_nodes(node):
+            nodes.append((child, new_ctx))
+        stack = nodes + stack
+
+        if len(nodes) == 0:
+            if len(stack) > 0:
+                _, next_ctx = stack[0]
+                while ctx != next_ctx:
+                    this_ctx = ctx.pop(0)
+                    if this_ctx in folding_starts:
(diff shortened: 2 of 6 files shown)


andfoy:
@ccordoba12, this one is ready for review

pyls/hookspecs.py
```diff
@@ -110,3 +110,8 @@ def pyls_settings(config):
 @hookspec(firstresult=True)
 def pyls_signature_help(config, workspace, document, position):
     pass
+
+
+@hookspec(firstresult=True)
+def pyls_folding_range(config, workspace, document):
```
gatesn: Can you move this alphabetically?

setup.py
```diff
@@ -92,6 +92,7 @@
             'rope_completion = pyls.plugins.rope_completion',
             'rope_rename = pyls.plugins.rope_rename',
             'yapf = pyls.plugins.yapf_format',
+            'folding = pyls.plugins.folding'
```
gatesn: Again, alphabetical

pyls/plugins/folding.py
```diff
@@ -0,0 +1,142 @@
...
+                    ctx, stack, line, col, folding_starts, folding_ends,
+                    tree_nodes, folding_ranges)
+
+    folding_ranges = sorted(folding_ranges)
+    ranges = __reduce_folding_ranges(folding_ranges)
+    return ranges
+
+
+@hookimpl
+def pyls_folding_range(document):
```
gatesn, line 16: As a style point, can you move this above the "helper" functions.

pyls/plugins/folding.py
```diff
@@ -0,0 +1,142 @@
...
+    program = str(document.source)
+    # Add an additional "pass" to ensure that there's always a next node
+    program = program + '\n\npass'
```
gatesn: Hah, clever

pyls/plugins/folding.py
```diff
@@ -0,0 +1,142 @@
...
+            offending_line = e.lineno
+            lines = program.splitlines(True)
+            lines = lines[:offending_line - 1]
+            program = ''.join(lines)
+            reparsing = True
+
+    # Parse again to add additional node
+    if reparsing:
+        program = program + '\n\npass'
+        tree = ast.parse(program)
```
gatesn: Should we recurse instead, to handle the case when there are multiple syntax errors?

test/plugins/test_folding.py
```diff
@@ -0,0 +1,129 @@
...
+                {'startLine': 39, 'endLine': 40},
+                {'startLine': 42, 'endLine': 43},
+                {'startLine': 45, 'endLine': 51},
+                {'startLine': 46, 'endLine': 47},
+                {'startLine': 48, 'endLine': 49},
+                {'startLine': 50, 'endLine': 51}]
+    assert ranges == expected
+
+
+def test_folding_syntax_error():
```
gatesn, line 148: Let's test multiple syntax errors?

andfoy:
@gatesn Do you agree in disabling code folding for syntax errors? We don't know how to handle this case with @ccordoba12

pyls/plugins/folding.py
```diff
@@ -0,0 +1,169 @@
...
+
+from pyls import hookimpl
+
+SKIP_NODES = (tree_nodes.Module, tree_nodes.IfStmt, tree_nodes.TryStmt)
+IDENTATION_REGEX = re.compile(r'(\s+).+')
+
+
+@hookimpl
+def pyls_folding_range(document):
+    program = str(document.source) + '\n'
```
ccordoba12: Is `str` really necessary here? I think it'd be better to write

    program = document.source + u'\n'
	
instead.

pyls/plugins/folding.py
```diff
@@ -0,0 +1,169 @@
...
+    last_operator = isinstance(last_leaf, tree_nodes.Operator)
+    node_is_operator = isinstance(node, tree_nodes.Operator)
+    last_operator = last_operator or not node_is_operator
+
+    end_line -= 1
+
+    modified = False
+    if isinstance(node.parent, tree_nodes.PythonNode):
+        kind = node.type
+        # print(f'Parent node: {kind}')
```
ccordoba12: Please remove this comment.

pyls/plugins/folding.py
```diff
@@ -0,0 +1,169 @@
+# pylint: disable=len-as-condition
+
+import re
+
+import parso
+import parso.python.tree as tree_nodes
+
+from pyls import hookimpl
+
+SKIP_NODES = (tree_nodes.Module, tree_nodes.IfStmt, tree_nodes.TryStmt)
```
ccordoba12, line 11: I think we should also provide folding on `if` and `try/except` statements. Why did you decide not to do that?

test/plugins/test_folding.py
```diff
@@ -0,0 +1,163 @@
+
```
ccordoba12, line 2: Palantir's copyright is missing here.

pyls/plugins/folding.py
```diff
@@ -0,0 +1,169 @@
+# pylint: disable=len-as-condition
```
ccordoba12, line 1: Please add Palantir's copyright above this line too.

ccordoba12 (commented):
@andfoy, I left some small comments for you, otherwise looks good.

pyls/plugins/folding.py
andfoy: VSCode does individual if/else folding

ccordoba12: I didn't know that. Then let's follow VSCode lead in this case.

PR #526, opened 2019-03-16
Fix #515 don't ignore LSP client's snippet support capabilities

Shougo:
Fix #515

Issue #515, opened 2019-03-05
pyls passes snippets to clients that don't support it


9ary:
Since #499, pyls passes snippets to the client even if it declares that it doesn't support it (with `capabilities.completion.completionItem.snippetSupport`), and the only way to work around this is to pass it [specific options](https://github.com/autozimu/LanguageClient-neovim/issues/737#issuecomment-465089402).

pyls should honor standard capabilities declared by the client.

Shougo:
I will fix the problem.  Please wait.

Shougo:
Done.

PR #671, opened 2019-10-11
If client declares supports, provide snippet completions by default

joaotavora:
This is the way python-language-server used to work before #519.  This
shouldn't conflict with the goals of #515 of #526, i.e. if the client
does _not_ declare support for snippets, none are sent.

* pyls/plugins/jedi_completion.py (pyls_completions): Default
should_include_params to True, like before.

### DIFF
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -49,7 +49,7 @@ def pyls_completions(config, document, position):
     snippet_support = completion_capabilities.get('completionItem', {}).get('snippetSupport')
 
     settings = config.plugin_settings('jedi_completion', document_path=document.path)
-    should_include_params = settings.get('include_params')
+    should_include_params = settings.get('include_params', True)
 
     return [_format_completion(d, snippet_support and should_include_params) for d in definitions] or None
 


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @joaotavora! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

joaotavora:
Signing many pages of agreement just for (re)adding a trivial `, True` isn't reasonable for me.  Take this PR as a suggestion.

ccordoba12:
> Take this PR as a suggestion.

Thanks but we're not interested in this.

joaotavora:
> Thanks but we're not interested in this.

Would you be so kind as to provide a reason? It's very much in vogue these days in the OSS world, providing a short explanation for why something isn't acceptable.

ccordoba12:
I think it's better for clients to declare if they want to use this feature instead of imposing it on them.

joaotavora:
> I think it's better for clients to declare if they want to use this feature instead of imposing it on them.

Thanks for getting back to me with an explanation.  I completely agree with you.  But if you read my pull request correctly, you will see that what you want to avoid is _not_ what my pull request is doing.

Is is true that before dfff3c965fea83c0429b17c37a0668b6023f519b, pyls was imposing snippets. but dfff3c965fea83c0429b17c37a0668b6023f519b went  a bit _too_ far.  

With my commit, if a client does not want to use the feature, it issues `snippetSupport: False`: the `pyls` code guarantees that snippets will not be imposed. But if it issues `snippetSupport: True`, then `pyls` should provide snippets without requiring that the user somehow additionally configures the server or Jedi to send snippets.

PR #677, opened 2019-10-15
Only use future library in Python 2

e-kwsm:
- install future only if python2
- remove futures which is not used any more


ccordoba12 (approved):
Looks good to me, thanks!

PR #679, opened 2019-10-20
Use flake8 utility instead of its internal API

### LINKED ISSUE #676: flake8 integration fails to ignore things that should be ignored
eg if I have 

```
[flake8]
ignore = W503,E203
```

in `~/.config/flake8`, those errors are reported by `pyls` anyway.

I think what's going on is that `flake8` only filters results when passing them to its formatter, round about [here](https://github.com/PyCQA/flake8/blob/e0612597094234cd7acaad3d64b6c7071cdf1831/src/flake8/style_guide.py#L573). But `pyls`'s `flake8` plugin looks at the unfiltered results.

youben11:
The usage of the internal API doesn't seem to work so well as the reports returned doesn't ignore the codes passed as arguments #676 . 

A solution around this was to re-implement flake8 linting via its utility.

Fixes #676.

### DIFF
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -98,8 +98,9 @@ def parse_stdout(document, stdout):
     """
 
     diagnostics = []
-    parsed_lines = re.findall(r'(.*):(\d*):(\d*): (\w*) (.*)', stdout)
-    for parsed_line in parsed_lines:
+    lines = stdout.splitlines()
+    for raw_line in lines:
+        parsed_line = re.findall(r'(.*):(\d*):(\d*): (\w*) (.*)', raw_line)
         if not parsed_line:
             continue
         _, line, character, code, msg = parsed_line
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -101,7 +101,8 @@ def parse_stdout(document, stdout):
     lines = stdout.splitlines()
     for raw_line in lines:
         parsed_line = re.findall(r'(.*):(\d*):(\d*): (\w*) (.*)', raw_line)
-        if not parsed_line:
+        if not parsed_line or len(parsed_line) != 5:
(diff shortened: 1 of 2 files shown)


youben11:
ping @ccordoba12

ccordoba12:
@youben11, any call to `flake8` will be made through `subprocess` from now on? That could create serious lags in big files.

Have you tested this to see how it works in a real life situation?

youben11:
I tested it against pyls biggest files with/without flake8 and it takes relatively the same time to get the diagnostics.

I thought about this before implementing it this way, and found out that pylint is using subprocess [under the hood](https://github.com/PyCQA/pylint/blob/master/pylint/epylint.py#L129) for a while without raising any issues.

pyls/plugins/flake8_lint.py
```diff
@@ -21,23 +22,64 @@ def pyls_lint(config, document):
...
+        cmd.extend(args)
+        p = Popen(cmd, stdout=PIPE, stderr=PIPE)
+    except IOError:
+        log.debug("Can't execute flake8. Trying with 'python -m flake8'")
+        cmd = ['python', '-m', 'flake8']
+        cmd.extend(args)
+        p = Popen(cmd, stdout=PIPE, stderr=PIPE)
+    stderr = p.stderr.read().decode()
+    if stderr:
+        log.error("Error while running flake8 '%s'", stderr)
```
ccordoba12, line 53: Should we `return` at this point? I mean, if there's an error here we shouldn't continue further, right?

ccordoba12 (commented):
@youben11, I left a small comment for you. Otherwise looks good to me.

pyls/plugins/flake8_lint.py
youben11: There might be cases where flake8 output some warnings to stderr while still doing the job, it's common in cli utilities so it's not really a sign that we shouldn't check flake8 stdout.

ccordoba12: Ok, thanks for the clarification.

PR #680, opened 2019-10-23
Add Jedi support for extra paths and different environment handling

goanpeca:
This adds the ability to pass more entries to sys.path or a different Python interpreter to Jedi to get additional or new completions (respectively) for them.

PR #682, opened 2019-10-23
Add completion tests for several libraries of the Scientific Python stack

ccordoba12:
This will allow us to verify that Jedi doesn't break completions for libraries like Numpy, Pandas and Matplotlib (the ones tested here).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,8 @@ setup(
         'pylint': ['pylint'],
         'rope': ['rope>0.10.5'],
         'yapf': ['yapf'],
-        'test': ['versioneer', 'pylint', 'pytest', 'mock', 'pytest-cov', 'coverage'],
+        'test': ['versioneer', 'pylint', 'pytest', 'mock', 'pytest-cov',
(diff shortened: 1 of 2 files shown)


ccordoba12:
@gatesn, this clearly shows that Jedi **0.15** broke completions for those libraries, which we discovered when we tried to update our Spyder dependency on the `python-language-server` to use version 0.29.

Therefore, I'm considering to push a new PR that adds back support for Jedi **0.14.1**, along with **0.15**, to have a `python-language-server` that works for Numpy, Pandas, etc. And on the Spyder side we'd demand **0.14.1** to have the version that we really need.

This is my plan because I think we can't demand Jedi **<0.15** here again due to 0.29+ being already packaged by several Linux distros (e.g. [Arch](https://www.archlinux.org/packages/community/any/python-language-server/)).

Please let me know if you think there's a better way to handle this very ugly situation. If I don't hear from you by tomorrow, I'll continue with this plan because it's critical for us to start using 0.29+ due to the PRs submitted by our team (#680, #665, #655 and more to come).

Issue #86, opened 2017-08-07
Add code actions to add/remove/sort imports


tomv564:
A sort imports action could always be available.
An add/remove import action could be registered for locations where pyflakes complained about missing/unused imports.

See: https://github.com/alecthomas/importmagic for a library supporting this functionality.

gatesn:
Feel free to steal and build off https://github.com/palantir/python-language-server/tree/ngates/auto-import if you have time.

It's a little too rough right now. Ideally we would want import magic to report the position it found the unimported modules instead of searching for them :/

ghost:
+1, import magic and isort combo would be amazing. I would probably leave isort as a separate manual command (same as YAPF).

ghost:
To clarify, having a separate isort command would allow the user to clean the imports manually, not only when new ones are added automatically.

ghost:
+1 for including importmagic! There are atom and vscode implementations already, guess it should be easy?

zewa666:
any updates on this one?

youben11:
@zewa666 implemented in #685. You can start testing if you'd like and report any issue that may arise.

PR #685, opened 2019-10-28
Code action to add/delete imports

youben11:
This is based on the previous work of @gatesn 

- Actions to import suggested modules based on the index of symbols of all available python modules (might be slow to build but cached for future use)
- Linting hint about unresolved symbols only as we aren't sure if those symbols aren't just variables.

#### Last features

- Actions to remove unreferenced imports
- Linting warning about unreferenced imports, variables and functions

related #86


youben11:
@gatesn is there any way to make the call to pyls_code_actions non blocking?

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,187 @@
...
+def _get_index(sys_path):
+    """Build index of symbols from python modules.
+    Cache the index so we don't build it multiple times unnecessarily.
+    """
+    key = tuple(sys_path)
+    if key not in _index_cache:
+        log.debug("Started building importmagic index")
+        index = importmagic.SymbolIndex()
+        # The build tend to be noisy
+        logging.getLogger('importmagic.index').setLevel(logging.ERROR)
```
gatesn: Think we should leave log levels up to the operator

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,187 @@
...
+_index_cache = {}
+
+
+def _get_index(sys_path):
+    """Build index of symbols from python modules.
+    Cache the index so we don't build it multiple times unnecessarily.
+    """
+    key = tuple(sys_path)
+    if key not in _index_cache:
+        log.debug("Started building importmagic index")
```
gatesn: Maybe an info level log?

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,187 @@
...
+            'severity': lsp.DiagnosticSeverity.Hint,
+        }
+
+    Args:
+        document: The document to be linted.
+    Returns:
+        A list of dictionaries.
+    """
+    scope = importmagic.Scope.from_source(document.source)
+    unresolved, _unreferenced = scope.find_unresolved_and_unreferenced_symbols()
```
gatesn: Maybe file a follow-up ticket to support a code action for removing unreferenced imports?

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,187 @@
...
+            if score < min_score:
+                # Skip low score results
+                continue
+
+            actions.append(generate_add_action(document, index, module, variable))
+
+    return actions
+
+
+def generate_add_action(document, index, module, variable):
```
gatesn: Prefix with `_` for a "private" function

gatesn (changes requested):

pyls/plugins/importmagic_lint.py
youben11: Was using them for seeing if the call is blocking and calculating the time of the index build, but yeah, an INFO level log would make more sense.

pyls/plugins/importmagic_lint.py
youben11: Nice feature to have, will work on it.

zewa666:
@youben11 I've checked out your version and tried it with a Webdriver/Selenium sample. Works fantastic to e.g autoimport `from selenium import webdriver`. What doesn't work though is importing local references from another file. Is that the supposed way how importmagic works?

youben11:
Hi @zewa666, thank you for testing this functionality and providing feedback.

Yeah the index of symbols is built from the python modules found under sys.path, but I will certainly add the local workspace so you can have suggestions to import from project files.

pyls/_utils.py
```diff
@@ -198,3 +199,7 @@ def is_process_alive(pid):
             return e.errno == errno.EPERM
         else:
             return True
+
+
+def camel_to_underscore(camelcase):
+    return re.sub('([A-Z]+)', r'_\1', camelcase).lower()
```
gatesn: I think this would rewrite `HelloWorld` to `_hello_world`. Can we be more specific, e.g. this: https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,264 @@
...
+ADD_IMPORT_COMMAND = 'importmagic.addimport'
+REMOVE_IMPORT_COMMAND = 'importmagic.removeimport'
+MAX_COMMANDS = 4
+UNRES_RE = re.compile(r"Unresolved import '(?P<unresolved>[\w.]+)'")
+UNREF_RE = re.compile(r"Unreferenced import '(?P<unreferenced>[\w.]+)'")
+
+_index_cache = {}
+
+
+def _get_index(sys_path):
```
gatesn: Let's make this return a future, that way we can skip results instead of blocking on indexing.

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,264 @@
...
+                continue
+
+            diagnostics.append({
+                'source': SOURCE,
+                'range': {
+                    'start': {'line': line_no, 'character': pos},
+                    'end': {'line': line_no, 'character': pos + len(unres)}
+                },
+                'message': "Unresolved import '%s'" % unres,
+                'severity': lsp.DiagnosticSeverity.Hint,
```
gatesn, line 182: Should this not be a warning / error? Or does it overlap with e.g. pyflakes results

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,264 @@
...
+    # Might be slow but is cached once built
+    # TODO (youben): add project path for indexing
+    index = _get_index(sys.path)
+    actions = []
+    diagnostics = context.get('diagnostics', [])
+    for diagnostic in diagnostics:
+        if diagnostic.get('source') != SOURCE:
+            continue
+        message = diagnostic.get('message', '')
+        if message.startswith('Unreferenced'):
```
gatesn: Why are we doing a `startswith` _and_ a regex match? Isn't one sufficient?

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,264 @@
...
+    min_score = conf.get('minScore', 1)
+    log.debug("Got importmagic settings: %s", conf)
+    importmagic.Imports.set_style(**{_utils.camel_to_underscore(k): v for k, v in conf.items()})
+
+    # Might be slow but is cached once built
+    # TODO (youben): add project path for indexing
+    index = _get_index(sys.path)
+    actions = []
+    diagnostics = context.get('diagnostics', [])
+    for diagnostic in diagnostics:
```
gatesn, line 232: I wonder if it's easier to just call `pyls_lint(document)`, then try to find the context diagnostic in the results. That way we don't have to parse the messages, just have to do an equality check.

e.g. `action_diags = [diag for diag in pyls_lint(document) if diag in context.get('diagnostics', [])]`

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,264 @@
...
+        },
+        'edits': [{
+            'range': {
+                'start': {'line': args['startLine'], 'character': 0},
+                'end': {'line': args['endLine'], 'character': 0},
+            },
+            'newText': args['newText']
+        }]
+    }]}
+    workspace.apply_edit(edit)
```
gatesn, line 327: I think newer versions of clients support putting the edit in the code action:

```
/**
 * A code action represents a change that can be performed in code, e.g. to fix a problem or
 * to refactor code.
 *
 * A CodeAction must set either `edit` and/or a `command`. If both are supplied, the `edit` is applied first, then the `command` is executed.
 */
export interface CodeAction {

	/**
	 * A short, human-readable, title for this code action.
	 */
	title: string;

	/**
	 * The kind of the code action.
	 *
	 * Used to filter code actions.
	 */
	kind?: CodeActionKind;

	/**
	 * The diagnostics that this code action resolves.
	 */
	diagnostics?: Diagnostic[];

	/**
	 * The workspace edit this code action performs.
	 */
	edit?: WorkspaceEdit;

	/**
	 * A command this code action executes. If a code action
	 * provides an edit and a command, first the edit is
	 * executed and then the command.
	 */
	command?: Command;
}
```

But I'm not sure which client capability we should check. What you've got now is probably safer

pyls/plugins/importmagic_lint.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def _add_command_title(variable, module):
+    if not variable:
+        return 'Import "%s"' % module
+    return 'Import "%s" from "%s"' % (variable, module)
+
+
+def _remove_command_title(import_name):
+    return 'Remove import of "%s"' % import_name
```
gatesn: Maybe `Remove unused import of %s`

test/test_utils.py
```diff
@@ -87,3 +87,9 @@ def test_clip_column():
     assert _utils.clip_column(2, ['123\n', '123'], 0) == 2
     assert _utils.clip_column(3, ['123\n', '123'], 0) == 3
     assert _utils.clip_column(4, ['123\n', '123'], 1) == 3
+
+
+def test_camel_to_underscore():
+    assert _utils.camel_to_underscore('camelCase') == 'camel_case'
+    assert _utils.camel_to_underscore('hangClosing') == 'hang_closing'
+    assert _utils.camel_to_underscore('ignore') == 'ignore'
```
gatesn, line 95: Can you test a few more cases, e.g. empty string, `UpperCamel`, etc.

vscode-client/package.json
```diff
@@ -35,6 +35,11 @@
                     },
                     "uniqueItems": true
                 },
+                "pyls.plugins.importmagic_lint.minScore": {
```
gatesn, line 43: Can you add the `pyls.plugins.importmagic_lint.enabled` option

gatesn (changes requested):

test/test_utils.py
youben11: I thought camelCase always starts with a lowercase but looks like their is SomeVariants.

pyls/plugins/importmagic_lint.py
youben11: This is a hint because we actually don't know if this is a module to be imported or just a non defined variable, so better not to trigger a false positive. The unreferenced symbols are flagged with Warning because we know in advance if the symbol is either an import or a varialbe/function and the error should be consistent.

pyls/plugins/importmagic_lint.py
youben11: True, should be cleaner now.

pyls/plugins/importmagic_lint.py
youben11: True that the diagnostics list can grow bigger than the local diagnostics and we will just pass most of them. Also we don't need to do the equality check if we call pyls_lint() as all the diagnostics will be processed.

youben11:
The importmagic doesn't return the location of unreferenced/unresolved symbols so we needed to look them up ourselves, the search logic was a simple naive .find() call which might return false positives (e.g 'os' is found in 'pos'). Next I will be looking on how to simply tokenize the source code and search among the tokens so we make sure the matching is correct.

gatesn:
@youben11 perhaps you could use Jedi's `names()` call instead of walking the AST yourself. It may help, haven't verified: https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.names

youben11:
@gatesn the disadvantage with jedi.names is that it does only return definitions, symbols such as function calls aren't returned. The implementation using tokenize is pretty straightforward, however, I have to deal with the two different version of python here as the function takes different parameters for py2 and py3.

#### Update
Actually there is a backward compatible function to use that works for both version.

gjabell:
Any motion on this PR? Would love to have this functionality in the language server!

weeman1337:
Hi there! Thanks for working on the import function :+1: 

After reading the comments it is not clear to me what is left to do. Since I really want this feature I would be happy to contribute - if I would know hat to do :D

ccordoba12:
This project is not maintained anymore. However, the community created a fork of it at

https://github.com/python-lsp/python-lsp-server/

So please submit this PR there if you want to see it finally merged.

Mte90:
If someone is looking for the same feature in the latest pylsp: https://github.com/python-lsp/python-lsp-server/pull/199

PR #687, opened 2019-10-31
Don't go to definition of builtin modules

jackwickham:
If you do Go To Definition (F12 in vscode) on a string, it opens the builtins.pyi file for python 3, because the definition is not being filtered properly.

This PR corrects that behaviour so that neither python 2 nor 3 will go to the builtins file.

This doesn't affect the behaviour with imported builtins - they still respect the configuration.

### DIFF
--- a/pyls/plugins/definition.py
+++ b/pyls/plugins/definition.py
@@ -21,5 +21,6 @@ def pyls_definitions(config, document, position):
             }
         }
         for d in definitions
-        if d.is_definition() and d.line is not None and d.column is not None and d.module_path is not None and not d.in_builtin_module()
+        if d.is_definition() and d.line is not None and d.column is not None and d.module_path is not None
+        and not d.in_builtin_module()
     ]
--- a/pyls/plugins/definition.py
+++ b/pyls/plugins/definition.py
@@ -21,5 +21,5 @@ def pyls_definitions(config, document, position):
             }
         }
         for d in definitions
-        if d.is_definition() and d.line is not None and d.column is not None and d.module_path is not None
+        if d.is_definition() and d.line is not None and d.column is not None and d.module_path is not None and not d.in_builtin_module()
     ]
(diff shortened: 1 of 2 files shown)


jackwickham:
If going to builtin definitions would be desirable for some people then it could be made configurable, but I'm not sure if that's necessary.

test/plugins/test_definitions.py
```diff
@@ -40,7 +40,7 @@ def test_builtin_definition(config):
 
     # No go-to def for builtins
     doc = Document(DOC_URI, DOC)
-    assert len(pyls_definitions(config, doc, cursor_pos)) == 1
+    assert [] == pyls_definitions(config, doc, cursor_pos)
```
gatesn: Should be `assert not pyls_definitions(config, doc, cursor_pos)`

pyls/plugins/definition.py
```diff
@@ -21,5 +21,6 @@ def pyls_definitions(config, document, position):
             }
         }
         for d in definitions
-        if d.is_definition() and d.line is not None and d.column is not None and d.module_path is not None
+        if d.is_definition() and d.line is not None and d.column is not None and
```
gatesn: Can you pull this out into a function? It doesn't read very clearly anymore

gatesn (changes requested):

pyls/plugins/definition.py
```diff
@@ -20,6 +20,14 @@ def pyls_definitions(config, document, position):
                 'end': {'line': d.line - 1, 'character': d.column + len(d.name)},
             }
         }
-        for d in definitions
-        if d.is_definition() and d.line is not None and d.column is not None and d.module_path is not None
+        for d in definitions if d.is_definition() and not_internal_definition(d)
     ]
+
+
+def not_internal_definition(definition):
```
gatesn: Prefix with `_` :)

gatesn (approved):

PR #692, opened 2019-11-09
Prevent extra folding regions for iterables and flow nodes

andfoy:
This PR prevents the computation of extra folding regions.


ccordoba12 (approved):
Thanks a lot @andfoy! Things are working really nice now!

PR #694, opened 2019-11-11
Detect multiline imports and don't show snippets for them

muffinmad:
Proposed changes in `pyls/plugins/jedi_completion.py`

`pyls_completions`: Check variables before calling function.
`use_snippets`: Include previous lines in parsed snippet if they ends with `\`;
Limit splits count by line number; Check token type instead of token code.

### DIFF
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -72,7 +72,7 @@ def use_snippets(document, position):
     line -= 1
     while line > -1:
         act_line = lines[line]
-        if act_line.rstrip()[-1] == '\\':
+        if act_line.rstrip().endswith('\\'):
             act_lines.insert(0, act_line)
             line -= 1
         else:
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -39,6 +39,9 @@ _TYPE_MAP = {
     'statement': lsp.CompletionItemKind.Keyword,
 }
 
+# Types of parso nodes for which snippet is not included in the completion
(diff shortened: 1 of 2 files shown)


ccordoba12:
@muffinmad, thanks a lot for the quick response!

@steff456, please review this one.

pyls/plugins/jedi_completion.py
```diff
@@ -51,25 +51,34 @@ def pyls_completions(config, document, position):
...
     settings = config.plugin_settings('jedi_completion', document_path=document.path)
     should_include_params = settings.get('include_params')
-    include_params = (use_snippets(document, position) and
-                      snippet_support and should_include_params)
+    include_params = (snippet_support and should_include_params and
+                      use_snippets(document, position))
     return [_format_completion(d, include_params) for d in definitions] or None
 
 
+_IMPORTS = ('import_name', 'import_from')
```
steff456: Please define this at the beginning of the file, before of after `_TYPE_MAP`

steff456 (changes requested):
I left a minor comment on the PR, it looks great!

muffinmad:
Besides requested changes, detection import in multi-statement lines is added. Although cases like
``` python
a = 1; from datetime import date
```
and
``` python
from datetime import date; a = date
```
are not commonly used, it's good to handle them correctly.

Please see updated PR.

Thanks!

pyls/plugins/jedi_completion.py
muffinmad: Done

steff456 (approved):
Thanks for this PR @muffinmad !

PR #696, opened 2019-11-14
Use ujson instead of regular json

mpanarin:
as it is faster

pylint checks use regular json. It is pretty slow and would be better if it used ujson instead
https://github.com/dignio/ultrajson
https://artem.krylysov.com/blog/2015/09/29/benchmark-python-json-libraries/

related to https://github.com/palantir/python-jsonrpc-server/pull/24

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,8 @@ setup(
         'backports.functools_lru_cache; python_version<"3.2"',
         'jedi>=0.14.1,<0.16',
         'python-jsonrpc-server>=0.1.0',
-        'pluggy'
+        'pluggy',
+        'ujson<=1.35'
     ],
 
     # List additional groups of dependencies here (e.g. development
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -37,7 +37,8 @@
         'backports.functools_lru_cache; python_version<"3.2"',
         'jedi>=0.14.1,<0.16',
         'python-jsonrpc-server>=0.1.0',
-        'pluggy'
+        'pluggy',
+        'ujson<=1.35'
```
ccordoba12, line 41: Why this condition?

ccordoba12 (commented):
I agree with this, but I'd like to know what @gatesn thinks about it.

setup.py
mpanarin: no real reason, just decided to softfreeze the version.

gatesn:
Sounds good to me 👍

PR #698, opened 2019-11-17
Disable Pylint plugin by default

ccordoba12:
- Pylint generates too much warnings without a configuration file. Users that rely on it probably know this, so it's better to let them enable the plugin in an opt-in manner.
- This follows the example set by the flake8 plugin too.

### DIFF
--- a/pyls/plugins/pylint_lint.py
+++ b/pyls/plugins/pylint_lint.py
@@ -147,6 +147,13 @@ def _build_pylint_flags(settings):
 
 
 @hookimpl
+def pyls_settings():
+    # Default pylint to disabled because it requires a config
+    # file to be useful.
+    return {'plugins': {'pylint': {'enabled': False, 'args': []}}}
+
+
+@hookimpl
 def pyls_lint(config, document, is_saved):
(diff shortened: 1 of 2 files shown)


zewa666:
meh ... would have been nice if the release notes had pointed out this breaking change @ccordoba12. E.g like [here](https://github.com/angular/angular/blob/master/CHANGELOG.md#breaking-changes). I know its still pre v1 but perhaps something similar could be adopted so that it's immediately clear for consumers that something needs the attention.

ccordoba12:
Release notes are posted in the Releases page, not in that file.

zewa666:
True, I just meant that the releases page entry could highlight breaking changes specifically as they might fall through otherwise. The link was just meant to show a sample, e.g a separate section per Release notes

AndreaOrru:
Even though this variable is set to nil, pylint sporadically runs. It's really annoying :(

AndreaOrru:
I found that Pylint may still be run by `flycheck` as the next checker in chain even when you use `flake8`.

I solved it by doing:
```
(add-to-list 'flycheck-disabled-checkers 'python-pylint)
```

PR #699, opened 2019-11-18
Show completions with and without params

muffinmad:
In some cases I don't want completion with snippets:

```python
class Foo(object):
    def __init__(self, a, b, c):
        pass

class Bar(Foo|
```

```python
from datetime import date, datetime

if foo:
    fn = date|
else:
    fn = datetime|
```

```python
def foo(a, b, c):
    pass

def bar(a, b, c):
    pass

z = foo(a, b, c)

# And now I need to replace `foo` with `bar`
z = bar|(a, b, c)
```

Assume cursor is at `|` and completions are invoked. Applying completion will insert snippet and I must delete it in order to have just class/function name.

IMO there are no way we can properly decide to include params or not.

My proposal is to include both candidates without and with params in completions list.

This PR add new option `include_params_both`. With `include_params` and `include_params_both` enabled completions list will include both variants.

There also side change: `label` is always equal to `insertText` in completion item. Is there any reason to completion label include all params?

### DIFF
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -43,11 +43,6 @@ _TYPE_MAP = {
 # Types of parso nodes for which snippet is not included in the completion
 _IMPORTS = ('import_name', 'import_from')
 
-# Provide completions with snippets
-_USE_PARAMS_OFF = 0
-_USE_PARAMS_ON = 1
-_USE_PARAMS_BOTH = 2
-
 
 @hookimpl
 def pyls_completions(config, document, position):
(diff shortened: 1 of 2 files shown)


ccordoba12:
This is an interesting idea, thanks @muffinmad!

mpanarin:
This is why I disabled this completion snippets completely.

Is it worth creating a different variable completely for this? As it won't make any sense to enable `include_params_both` without the original one.
Maybe some kind of switch?
like:
falsy val - disabled
truthy val - enabled with only params (for backwards compat)
"both" string specifically - enable both

muffinmad:
> Is it worth creating a different variable completely for this? As it won't make any sense to enable `include_params_both` without the original one.
> Maybe some kind of switch?
> like:
> falsy val - disabled
> truthy val - enabled with only params (for backwards compat)
> "both" string specifically - enable both

At first I was going to use `both` value of the `include_params` variable for this. But in https://github.com/palantir/python-language-server/blob/45702ec1d0f02834e0f076f683dd0f4bb6ba14d6/vscode-client/package.json#L53 `include_params` described as `bool` so I'm not sure if string is allowed.

Maybe `int` will fit? Like:
- `0` - no snippets
- `1` - snippets only
- `2` - both

Or both variants of completion can be used if `include_params` is enabled. Having only snippets in completion is little annoying.

muffinmad:
Having only snippets in completion is little annoying indeed so both variants of completion is used now when `include_params` is set.

PR #700, opened 2019-11-18
Add `--config` param for flake8

mpanarin:
Allow explicit pass of config to flake. This can be useful for weird per-project configs that are not in default directories.
 
@gatesn is this config (https://github.com/palantir/python-language-server/blob/develop/pyls/config/flake8_conf.py#L32) used somewhere?
I was unable to find where :\

### DIFF
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -1,6 +1,7 @@
 # Copyright 2019 Palantir Technologies, Inc.
 """Linter pluging for flake8"""
 import logging
+from os import path
 import re
 from subprocess import Popen, PIPE
 from pyls import hookimpl, lsp
@@ -20,6 +21,7 @@ def pyls_lint(config, document):
     log.debug("Got flake8 settings: %s", settings)
 
     opts = {
+        'config': settings.get('config'),
         'exclude': settings.get('exclude'),
         'filename': settings.get('filename'),
         'hang-closing': settings.get('hangClosing'),
@@ -27,6 +29,14 @@ def pyls_lint(config, document):
         'max-line-length': settings.get('maxLineLength'),
         'select': settings.get('select'),
     }
+
+    # flake takes only absolute path to the config. So we should check and
+    # convert if necessary
+    if opts.get('config') and not path.isabs(opts.get('config')):
+        opts['config'] = path.abspath(path.expanduser(path.expandvars(
+            opts.get('config')
+        )))
+        log.debug("using flake8 with config: %s", opts['config'])
 
     # Call the flake8 utility then parse diagnostics from stdout
     args = build_args(opts, document.path)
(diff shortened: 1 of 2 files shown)


ccordoba12:
@youben11, could you take a look at this one? Thanks!

youben11:
@mpanarin flake8_conf.py isn't used by default, but if it's mentioned [here](https://github.com/palantir/python-language-server/blob/develop/pyls/config/config.py#L16) then it would load flake8 conf file on startup. Could also be configured via vscode client.

I think this PR is about specifying a configuration file to use after loading the project configuration file (.flake8, setup.cfg, tox.ini), am I right? Because the 'config' option here would be parsed from one of the project config file to be loaded as 'plugins.flake8.config'

mpanarin:
@youben11 yes, I saw that it is imported there. I just don't understand how it is used further down with flake8 tool. Especially as `flake8` is now called with Popen, instead of an api. Are you sure that is not a dead code?

Anyway, this pr is about specifying the config file to the cli tool, yes. 
Its WIP, I need to check a few thing

youben11:
@mpanarin flake8_config will load config as 'plugins.flake8.*' then the flake8 plugin will use them for linting, you can see [here](https://github.com/palantir/python-language-server/blob/develop/pyls/plugins/flake8_lint.py#L18) how config values are used, this function will be called each time a lint operation is to be done (if the plugin is enabled).

I may be wrong about the usecase, but I wonder if we should parse that option from pyls settings (that can be configured by clients) instead of parsing that from a config file?

mpanarin:
> I may be wrong about the usecase, but I wonder if we should parse that option from pyls settings (that can be configured by clients) instead of parsing that from a config file?

That is exactly what I intended to do. I updated the PR and removed the param from flake8_config. Checked for it to work in emacs environment. 

Should I add tests for this pretty small change?
Also, I am not sure where to note, that other parameters such as `max-line-length` overwrite params from config. Should it be noted on the specific frontend insteaad? (VScode/emacs/etc.)

youben11:
I think a quick test for ensuring the config option is working as intended would be great.

@ccordoba12 I'm note sure about parameters documentation.

mpanarin:
@youben11 @ccordoba12 updated

pyls/plugins/flake8_lint.py
```diff
@@ -64,13 +74,15 @@ def build_args(options, doc_path):
...
         arg = None
         if isinstance(arg_val, list):
             arg = '--{}={}'.format(arg_name, ','.join(arg_val))
         elif isinstance(arg_val, bool):
             if arg_val:
                 arg = '--{}'.format(arg_name)
-        elif isinstance(arg_val, int):
+        else:
             arg = '--{}={}'.format(arg_name, arg_val)
         if arg:
```
youben11: I think we should get rid of this check as you made sure `arg` always holds a value

youben11 (changes requested):

pyls/plugins/flake8_lint.py
mpanarin: updated

youben11 (approved):

mpanarin:
@ccordoba12 Hello,
PR is ready and reviews are attended, can this be merged, please?)

ccordoba12:
Sorry, since I saw it marked as WIP I thought it was not ready.

PR #701, opened 2019-11-19
PR: Fix snippet parsing when a slash character is present

steff456:
Fixes the syntax of the snippet when a slash (`/`) character is present.

Fixes https://github.com/spyder-ide/spyder/issues/10761

snippet

PR #703, opened 2019-11-20
Don't include params notations in snippet

muffinmad:
Related PR: https://github.com/palantir/python-language-server/pull/701

`/` and `*` in parameters list are not parameters but notations. See [PEP570](https://www.python.org/dev/peps/pep-0570/#syntax-and-semantics).


ccordoba12:

PR #704, opened 2019-11-21
Remove ujson dependency on Windows (requires MSVC compiler)

### LINKED ISSUE #10763: Fails to install Spyder4.0.0rc2 on Windows due to ujson dependency
## Problem Description
First of all, this is a dependency issue so it may be not solved by the Spyder-team but I thought that you may want to know about it.

Spyder 4.0.0rc2 actually fails to install (via pip) on Windows if MSVC compiler is not installed. This is related to the fact that some dependencies depend on `ujson` package, which require MSVC 14 as shown in the log below.

So the dependency tree goes like that: rc2 requires `python-language-server` 0.31.0, which requires `ujson` and `python-jsonrpc-server` 0.3.0, which requires `ujson` too:
```
spyder 4.0.0rc2
    python-language-server 0.31.0
        ujson (since python-language-server 0.31.0, previous 0.30.0)
        python-jsonrpc-server 0.3.0
            ujson (since python-jsonrpc-server 0.3.0, previous 0.2.0)
```

This is working with rc1 when pre-installing first python-jsonrpc-server v0.2.0, then python-language-server 0.30.0.

### What steps reproduce the problem?

1. Install release candidate 2 on W10  `pip install --pre spyder`

### Error

```
[...]
  Running setup.py install for ujson ... error
[...]
    Warning: 'classifiers' should be a list, got type 'filter'
    running install
    running build
    running build_ext
    building 'ujson' extension
    error: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++ Build Tools": https://visualstudio.microsoft.com/downloads/
    ----------------------------------------
```

## Versions

* Spyder version: -
* Python version: 3.7.5
* Qt version: -
* PyQt version: - 
* Operating System name/version: Windows 10, AMD64

dalthviz:
As discussed with @ccordoba12
Fixes spyder-ide/spyder#10763

Depends on  palantir/python-jsonrpc-server#26 and a new release of `python-jsonrpc-server`

### TODO:
 - [x] Bump up min version required for `python-jsonrpc-server`
 - [x] Fix tests

### DIFF
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -6,7 +6,7 @@ import sys
 
 try:
     import ujson as json
-except Exception:
+except ImportError:
     import json
 
 from .python_ls import (PythonLanguageServer, start_io_lang_server,
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -6,7 +6,7 @@ import sys
 
 try:
     import ujson as json
-except ImportError:
+except Exception:  # pylint: disable=broad-except
     import json
 
 from .python_ls import (PythonLanguageServer, start_io_lang_server,
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -4,7 +4,10 @@ import logging
 import logging.config
 import sys
 
-import ujson as json
+try:
+    import ujson as json
+except Exception:
+    import json
 
 from .python_ls import (PythonLanguageServer, start_io_lang_server,
                         start_tcp_lang_server)
--- a/setup.py
+++ b/setup.py
@@ -36,9 +36,9 @@ setup(
         'future>=0.14.0; python_version<"3"',
         'backports.functools_lru_cache; python_version<"3.2"',
         'jedi>=0.14.1,<0.16',
-        'python-jsonrpc-server>=0.3.0',
+        'python-jsonrpc-server>=0.3.1',
         'pluggy',
-        'ujson<=1.35; platform_system=="Linux"'
+        'ujson<=1.35; platform_system!="Windows"'
     ],
 
     # List additional groups of dependencies here (e.g. development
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -38,7 +38,7 @@
         'jedi>=0.14.1,<0.16',
         'python-jsonrpc-server>=0.3.0',
         'pluggy',
-        'ujson<=1.35'
+        'ujson<=1.35; platform_system=="Linux"'
```
mpanarin: I am not an expert here, but will this prevent mac users form using ujson as well? 
Something like `platform_system != "Windows"`(pseudo, as I am not sure how to properly select it) maybe?

mpanarin: same for `jsonrpc` package (where it is actually more important than here)

dalthviz: Yep you're right @mpanarin it should be changed for either `platform_system != "Windows"` or something like `platform_system == "Linux" or ´platform_system == "Darwin"` (here and over palantir/python-jsonrpc-server#26). Thanks for the feedback!

ccordoba12: @dalthviz, I agree with @mpanarin. Please change this condition to `platform_system!="Windows"`

ccordoba12:
@dalthviz, please also fix the pylint warning generated in the Python 2 build of Circle.

pyls/__main__.py
```diff
@@ -4,7 +4,10 @@
 import logging.config
 import sys
 
-import ujson as json
+try:
+    import ujson as json
+except ImportError:
```
ccordoba12: I think we should use `Exception` here instead of `ImportError` to catch any possible problem when importing `ujson`.

ccordoba12 (approved):
Thanks @dalthviz!

PR #711, opened 2019-12-04
Fix TCPServer being bound before allow_reuse_address is set

bmwalters:
Currently, `server_bind` is [being called](https://github.com/python/cpython/blob/3.8/Lib/socketserver.py#L452) before `allow_reuse_address` is set on the `TCPServer` instance. This leads to the flag having no effect.

This patch adjusts the code to call `server_bind` only after setting the flag.

See the bottom of [this stackoverflow answer](https://stackoverflow.com/a/42147927) for more context.

### DIFF
--- a/pyls/python_ls.py
+++ b/pyls/python_ls.py
@@ -68,10 +68,12 @@ def start_tcp_lang_server(bind_addr, port, check_parent_process, handler_class):
          'SHUTDOWN_CALL': shutdown_server}
     )
 
-    server = socketserver.TCPServer((bind_addr, port), wrapper_class)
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @bmwalters! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12:
@bmwalters, please sign the CLA so I can merge this one.

ccordoba12 (approved):
Thanks @bmwalters for your help with this one!

PR #719, opened 2019-12-20
Address autopep8 overriding pycodestyle's continued_indentation

johnhany97:
### Before this PR
Given a Python file like:
```
def hello():
    print("Python", ("Hello",
                                     "World"))
```
The diagnostics report of pycodestyle is:
```
[{'source': 'pycodestyle', 'range': {'start': {'line': 2, 'character': 37}, 'end': {'line': 2, 'character': 47}}, 'message': 'E127 21', 'code': 'E127', 'severity': 2}]
```
### Root cause & Description
While the error code, `E127`, is expected, this is not the expected error message.
This is because `autopep8` overrides `pycodestyle`'s implementation of `continued_indentation` which is the check used to determine this error code, among others. `autopep8`'s implementation however returns a different thing than what we expect out of the `pycodestyle` implementation, leading to a scenario where the diagnostics report contains an unuseful error message.

### After this PR
The output is as expected:
```
[{'source': 'pycodestyle', 'range': {'start': {'line': 2, 'character': 37}, 'end': {'line': 2, 'character': 47}}, 'message': 'E127 continuation line over-indented for visual indent', 'code': 'E127', 'severity': 2}]
```

### DIFF
--- a/pyls/plugins/pycodestyle_lint.py
+++ b/pyls/plugins/pycodestyle_lint.py
@@ -2,6 +2,11 @@
 import logging
 import pycodestyle
 from pyls import hookimpl, lsp
+from autopep8 import continued_indentation as autopep8_c_i
+
+if autopep8_c_i in pycodestyle._checks['logical_line']:
+    del pycodestyle._checks['logical_line'][autopep8_c_i]
+    pycodestyle.register_check(pycodestyle.continued_indentation)
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


pyls/plugins/pycodestyle_lint.py
```diff
@@ -1,8 +1,13 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import logging
 import pycodestyle
+from autopep8 import continued_indentation as autopep8_c_i
 from pyls import hookimpl, lsp
 
+if autopep8_c_i in pycodestyle._checks['logical_line']:
```
ccordoba12, line 10: Please add a comment above this line to explain the reasoning behind this block.

johnhany97: Gladly 👍

ccordoba12 (approved):
Thanks @johnhany97!

afflux:
Unfortunately, this change breaks formatting when a document has continuation lines.

Try formatting this:
```python
def foo():
    print('asdf',
    file=None
    )
```

PR #728, opened 2020-01-11
Add option to activate fuzzy completions in Jedi

### LINKED ISSUE #718: Add support for fuzzy completions in Jedi
Jedi added fuzzy completions \o/

Probably worth updating the dependency, plus make a var for enabling/disabling fuzzy completion
https://github.com/davidhalter/jedi/issues/1063#issuecomment-567887861

benmezger:
This adds fuzzy support to Jedi as requested by #718. Since fuzzy is only enabled in 0.15.2 (see https://github.com/davidhalter/jedi/blob/master/CHANGELOG.rst#0152-2019-12-20), this PR also update's Jedi's requirements.

Fixes #718


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @benmezger! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12:
> this PR also update's Jedi's requirements.

Please don't do that. Instead, please make things work for older Jedi versions as well.

pyls/plugins/jedi_completion.py
```diff
@@ -45,8 +45,9 @@
 
 @hookimpl
 def pyls_completions(config, document, position):
+    settings = config.plugin_settings('jedi_completion', document_path=document.path)
     try:
-        definitions = document.jedi_script(position).completions()
+        definitions = document.jedi_script(position).completions(fuzzy=settings.get("fuzzy", False))
```
mpanarin: you probably can check the version here with `jedi.__version__`

Although I am not sure what's the issue with bumping the jedi version for the release

benmezger: Maybe bumping Jedi's version is a bit.. rude, but I agree that bumping Jedi's should be done sooner or later.

Either way, i've updated the code to make sure we first check Jedi's version before enabling fuzzy completion and removed Jedi's version bump, see: 3ecff5f18854d6e6459184ef054d909029420c3b and 7e6d27a

benmezger:
@ccordoba12 I believe using `LooseVersion` should do the job, but let me know if there is anymore improvements or if I missed anything.

pyls/plugins/jedi_completion.py
```diff
@@ -1,10 +1,17 @@
...
+from distutils.version import LooseVersion
 import parso
+
+import jedi
 from pyls import hookimpl, lsp, _utils
 
+
 log = logging.getLogger(__name__)
 
+JEDI_VERSION = jedi.__version__
```
ccordoba12: This constant is defined here

https://github.com/palantir/python-language-server/blob/19b10c47988df504872a4fe07c421b0555b3127e/pyls/_utils.py#L13

so please that one instead.

benmezger: See: 7c02479

vscode-client/package.json
```diff
@@ -55,6 +55,11 @@
                     "default": true,
                     "description": "Auto-completes methods and classes with tabstops for each parameter."
                 },
+                "pyls.plugins.jedi_completion.fuzzy": {
+                    "type": "boolean",
+                    "default": true,
```
ccordoba12: Please change this to `false`, because the setting is false by default, right?

davidhalter:
> Please don't do that. Instead, please make things work for older Jedi versions as well.

@ccordoba12 Just out of interest, why do you want to support older versions of Jedi? IMO that's just a waste of resources, since you could force users to install a new one :).

ccordoba12:
> Just out of interest, why do you want to support older versions of Jedi?

Because some Jedi versions (e.g. 0.15.0 and 0.15.1) broke completion for scientific Python libraries, and we had to do some ugly workarounds to support older versions that still work with them.

I really don't want my comment to sound as a criticism on your work as a maintainer (I'm also a maintainer, so I understand how hard that job is), but I don't want to be hit by that kind of bugs for now.

By the way, we're planning to migrate soon to use Jedi 0.16, so this is just a temporary measure.

davidhalter:
> Because some Jedi versions (e.g. 0.15.0 and 0.15.1) broke completion for scientific Python libraries, and we had to do some ugly workarounds to support older versions that still work with them.

While I'm sorry that this happened, in the future you should probably just not upgrade and try to open issues in Jedi. Jedi is probably soon going to be feature complete and 1.0 is coming.

vscode-client/package.json
benmezger: @ccordoba12 see affc871

ccordoba12:
@benmezger, please merge this PR with master and fix the merge conflicts.

benmezger:
@ccordoba12 ✅

ccordoba12 (approved):
Thanks for this @benmezger! And thanks for your patience too.

PR #729, opened 2020-01-11
Fix: Run pytest in parallel using xdist

benmezger:
This runs pytest in parallel by using xdist instead. Pytest takes around ~64.88s  in my machine, it now takes around ~41.23s, which is a plus when running tests within CircleCI

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,5 +14,6 @@ exclude = test/plugins/.ropeproject,test/.ropeproject
 [tool:pytest]
 testpaths = test
 addopts =
+    -n 4
     --cov-report html --cov-report term --junitxml=pytest.xml
     --cov pyls --cov test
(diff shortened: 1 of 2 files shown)


benmezger:
Hm, this is failing on `mypy`. I am unable to reproduce this on my machine, any ideas why this fails?

ccordoba12:
I don't think Jedi tests could be run in parallel, so this should be avoided.

benmezger:
@ccordoba12 this is a bit odd, as I am able to pass all tests on my machine (OSX). Why is `parso/cache.py:119` raising an `EOFError`?

ccordoba12:
Probably several processes can't access that file at the same time on Windows.

goanpeca:
xdist may have issues on win, so maybe just use it for non windows tests?

benmezger:
> xdist may have issues on win, so maybe just use it for non windows tests?

I haven't found any reference of getting xdist to not run parallel tests on Windows. Anything specific I should be looking for?

PR #731, opened 2020-01-17
Upload wheels first to try to get pypi to detect deps

gatesn:
As per https://github.com/boto/boto3/issues/2111


techtonik:
@gatesn did that help?

PR #734, opened 2020-01-21
Only add snippet completions when positional args are available

### LINKED ISSUE #11356: Completion: pyls improvements for snippets handling
* [x] Don't show snippets for single argument
* [x] Don't complete error classes with arguments or parenthesis
* [x] Don´t show snippets for multi-line import statements

Note: To be done over the pyls side

As discussed with @ccordoba12

dalthviz:
* Prevents the creation of snippets completions with none or single arguments like `..()$0` or `..(${1:args})$0`.

* Prevents the creation of snippets completions for multi-line imports statements like:
```python
from collections import (
    defaultdict)
```

Fixes spyder-ide/spyder#11356

### DIFF
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -84,6 +84,7 @@ def is_exception_class(name):
         # class definition in Python 2
         return False
 
+
 def use_snippets(document, position):
     """
     Determine if it's necessary to return snippets in code completions.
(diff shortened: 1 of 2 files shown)


dalthviz:
I think this is ready @ccordoba12 :+1:

pyls/plugins/jedi_completion.py
```diff
@@ -103,16 +112,17 @@ def _format_completion(d, include_params=True):
...
-        completion['insertTextFormat'] = lsp.InsertTextFormat.Snippet
-        snippet = d.name + '('
-        for i, param in enumerate(positional_args):
-            name = param.name if param.name != '/' else '\\/'
-            snippet += '${%s:%s}' % (i + 1, name)
-            if i < len(positional_args) - 1:
-                snippet += ', '
-        snippet += ')$0'
-        completion['insertText'] = snippet
+        if len(positional_args) > 1:
```
ccordoba12, line 137: I think we should insert `()` and put the cursor in the middle of them, in case `positional_args == 1`.

ccordoba12: For `positional_args == 0` we need to put the cursor to the right of `()`

ccordoba12: For `positional_args == 0` we need to put the cursor to the right of `()`

ccordoba12 (approved):
Thanks @dalthviz!

PR #740, opened 2020-02-03
Fix issues with parsing except's while computing folding

### LINKED ISSUE #11423: Code folding does not fold correctly when there is "except"
## Problem Description
When "except" in a function, the part of "except" codes are folded and a bug occurs.
code like this:

When try to fold the part of "except" codes, funB is folded unexpectedly.

## Versions
* Spyder version: 4.0.1
* Python version: 3.6.9
* Qt version: 5.9.7
* PyQt version: 5.9.2
* Operating System name/version: windows 7

andfoy:
Fixes spyder-ide/spyder#11423

### DIFF
--- a/pyls/plugins/folding.py
+++ b/pyls/plugins/folding.py
@@ -110,34 +110,33 @@ def __check_if_node_is_valid(node):
     return valid
 
 
+def __handle_skip(stack, skip):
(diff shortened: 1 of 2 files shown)


ccordoba12 (approved):
Thanks @andfoy!

Issue #737, opened 2020-01-28
Python2 support


mpanarin:
Python2 reached its EOL. Will `pyls` continue to support it? 

[Jedi already dropped support for python2](https://github.com/davidhalter/jedi/issues/1063#issuecomment-578457199) in its latest release. Also added a bunch of sweet improvements like:
- `Tensorflow, Numpy and Pandas completions should now be about 4-10x faster` 
- `Dict key completions are working now. e.g. d = {1000: 3}; d[10 will
expand to 1000`

pretty sure spyder community will benefit from this changes a lot.
On the other hand we have `rope` as a dependency that supports python3 only partly. But using it for completion is inferior to Jedi, imo. And renaming with it is super unstable with a big project, so I disabled rope completely. Maybe it is worth investing into making refactoring with jedi possible?
So rope could be dropped as a dependency

@ccordoba12 @gatesn

gatesn:
Agree with regard to Rope. I believe Jedi was looking into providing basic refactoring support but I might be wrong.

Seems like we might also be ready to drop Python 2 support from master but I’d like to keep a release branch around for hot fixes.

ccordoba12:
> pretty sure spyder community will benefit from this changes a lot.

Yes, totally! Thanks for bringing this up @mpanarin!

> Seems like we might also be ready to drop Python 2 support from master but I’d like to keep a release branch around for hot fixes.

Agreed, and that's what I was about to propose too: creating a `0.x` branch for bug fixes only and leaving `develop` to switch to Python 3, with the next release being `1.0.0`. I know that's an important milestone, but from the Spyder side we're committed to keep maintaining this project and make it as robust and stable as possible.

I'd also like to have a last release (0.32.0) before moving with this plan, with a couple of PRs that are ready to be merged.

@gatesn, what do you think?

gatesn:
Sounds good to me, let’s 1.0!

ccordoba12:
Glad you agreed! I'll release 0.32.0 and create the 0.x maintenance branch this weekend.

mcepl:
> Agree with regard to Rope. I believe Jedi was looking into providing basic refactoring support but I might be wrong.

This is a rope maintainer speaking. We are claiming now to support Python 3 (and at least our rather extensive test suite is passing), so if you can file any bugs in that regard to our issue tracker (https://github.com/python-rope/rope/issues), it would be helpful.

benmezger:
[...]

PR #748, opened 2020-02-23
Log local timezone

e-kwsm:
https://docs.python.org/3/library/logging.html#logging.Formatter.formatTime

### DIFF
--- a/pyls/__main__.py
+++ b/pyls/__main__.py
@@ -3,6 +3,7 @@ import argparse
 import logging
 import logging.config
 import sys
+import time
 
 try:
     import ujson as json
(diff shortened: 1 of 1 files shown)


goanpeca:
The server could be running on a different Time Zone, Why not add the correct one, instead of removing it?

e-kwsm:
CI fails since python 2 does not offer `time.localtime().tm_zone`, but support will be dropped (#737).

ccordoba12:
> CI fails since python 2 does not offer time.localtime().tm_zone, but support will be dropped

Ok, no problem. I'll leave this for 1.0, which will drop support for Python 2.

PR #749, opened 2020-02-24
Fix m_workspace__did_change_workspace_folders to follow the spec

zerocewl:
Changed signature for `m_workspace__did_change_workspace_folders` according to microsofts language-server-protocol [`DidChangeWorkspaceFolders`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeWorkspaceFolders).

see #720

Issue #720, opened 2019-12-21
Workspace folders not implemented completely


tomv564:
The signature for didChangeWorkspaceFolders

```python
def m_workspace__did_change_workspace_folders(self, added=None, removed=None, **_kwargs):
```

assumes `added` and `removed` are children of the `params` field, but according to the spec they should be nested under `event`? (See https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#workspace_didChangeWorkspaceFolders)

Results in: 
```
:: --> pyls workspace/didChangeWorkspaceFolders: {'event': {'removed': [], 'added': []}}
pyls: 2019-12-19 18:50:04,738 UTC - ERROR - pyls_jsonrpc.endpoint - Failed to handle notification workspace/didChangeWorkspaceFolders: {'event': {'removed': [], 'added': []}}
pyls: Traceback (most recent call last):
pyls: File "/usr/local/lib/python3.7/site-packages/pyls_jsonrpc/endpoint.py", line 142, in _handle_notification
pyls: handler_result = handler(params)
pyls: File "/usr/local/lib/python3.7/site-packages/pyls_jsonrpc/dispatchers.py", line 23, in handler
pyls: return method(**(params or {}))
pyls: File "/usr/local/lib/python3.7/site-packages/pyls/python_ls.py", line 363, in m_workspace__did_change_workspace_folders
pyls: for removed_info in removed:
pyls: TypeError: 'NoneType' object is not iterable
```

The server also does not take the initial list of workspaces from `initialize` - it only creates a single workspace for `rootUri`. Sending a `didChangeWorkspaceFolders` with an entry in `removed` that is not the _rootUri_ will result in a KeyError.

ccordoba12:
@andfoy, is this correct?

GNSubrahmanyam:
I am also getting this issue
```bash
python2    | python2 Server: 2020-04-24 18:01:25,246 UTC - ERROR - pyls_jsonrpc.endpoint - Failed to handle notification workspace/didChangeWorkspaceFolders: {u'event': {u'removed': [{u'uri': u'file:///workspace/', u'name': u'file:///workspace/'}], u'added': [{u'uri': u'file:///workspace/', u'name': u'file:///workspace/'}]}}
python2    | Traceback (most recent call last):
python2    |   File "/usr/local/lib/python2.7/dist-packages/pyls_jsonrpc/endpoint.py", line 142, in _handle_notification
python2    |     handler_result = handler(params)
[...]

PR #750, opened 2020-02-24
Fixed keyError if a workspace uri is wrong for a removed workspace

zerocewl:
Fixed keyError if a workspace uri is wrong for a removed workspace in `m_workspace__did_change_workspace_folders` + added simple Test.

Discovered in #720 

Rebased on #749 / works only on top of #749


ccordoba12:
Closing because this is included in #754.

Issue #751, opened 2020-02-24
flake8 plugin using fragile stderr.read() call


rdugan:
I'm using a combination of emacs, poetry, and pyls/flake8, and seeing instances where expected flake8 diagnostics are missing from the pyls response.  I've traced this back to a hanging stderr.read() call in the run_flake8() function:

https://github.com/palantir/python-language-server/blob/f6df42cae186e5389e0332e9b8406bb24bd84d3e/pyls/plugins/flake8_lint.py#L61

According to the [Python docs](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.stdin),  (scroll up a couple lines,) `Popen.communicate()` should be used rather than `Popen.stderr.read()` to avoid deadlocks due to "other OS pipe buffers filling up and blocking the child process" which appears to be what's happening in my situation.

PR #752, opened 2020-02-24
Switch to Popen.communicate() to avoid IO deadlocks in flake8 plugin

rdugan:
Fix to address https://github.com/palantir/python-language-server/issues/751

### DIFF
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -58,11 +58,10 @@ def run_flake8(args):
         cmd = ['python', '-m', 'flake8']
         cmd.extend(args)
         p = Popen(cmd, stdout=PIPE, stderr=PIPE)
-    stderr = p.stderr.read().decode()
+    (stdout, stderr) = p.communicate()
     if stderr:
-        log.error("Error while running flake8 '%s'", stderr)
-    stdout = p.stdout
-    return stdout.read().decode()
+        log.error("Error while running flake8 '%s'", stderr.decode())
+    return stdout.decode()
 
 
 def build_args(options, doc_path):


ccordoba12:
@youben11, please review this one.

goanpeca (approved):
Thanks for the fix @rdugan !

I guess writing a regression test for this might be tricky. Any ideas?

rdugan:
No problem...  It's a pretty minor change, which I thought it would just be a quick drop in replacement for the existing functionality (and covered by existing tests).  The test failures were unexpected, unfortunately I didn't have time to investigate yet.

goanpeca:
@rdugan You need to update the test that is failing. The test was doing some mocking and the mocking now fails because the mocked communicate call is not returning something that makes sense.

Something quick and dirty like:

```python
def test_flake8_config_param(config):

    class PopenMock:
        def __init__(self, *args, **kwargs):
            PopenMock.args = args

        def communicate(self):
            return bytes(), bytes()

    with patch('pyls.plugins.flake8_lint.Popen', new=PopenMock) as popen_mock:
        flake8_conf = '/tmp/some.cfg'
        config.update({'plugins': {'flake8': {'config': flake8_conf}}})
        _name, doc = temp_document(DOC)

        flake8_lint.pyls_lint(config, doc)
        call_args = popen_mock.args[0]

        assert 'flake8' in call_args
        assert '--config={}'.format(flake8_conf) in call_args

```

Might work!

---

On another note, you made this PR from your develop branch to the develop branch, and you should have created a separate branch first. 

```bash
git checkout develop
git checkout -b branch-with-this-fix  # silly example name
# Changes, git add git commit
git push branch-with-this-fix  # silly example name
```

So it is better if you create a new PR like this.

goanpeca (changes requested):
See the last comment!

rdugan:
Replaced w/ PR #757

PR #754, opened 2020-02-25
Follow LSP protocol when workspace folders are changed and adapt root workspace uri if folder is changed

### LINKED ISSUE #720: Workspace folders not implemented completely

zerocewl:
If a workspace folder is changed via `m_workspace__did_change_workspace_folders` and the **removed uri equals the `root uri`,** the `root_uri` will be adapted. Additional the `self.workspace` will be reseted with a new Workspace which contains the workspace.

The adoption of the `root_uri` is necessary if a the root workspace folder changes and the jedi completion should use this new workspace folder.

Fixes #720


ccordoba12:
@zerocewl, sorry for the big delay in taking a look at this one. Does this supersede your PRs in #749 and #750 completely? I mean, is this the only PR we need to review?

ccordoba12:

andfoy (approved):

PR #756, opened 2020-02-25
Create Spark Cost Checker

svanhooser:
- Adds a new linter that returns information when costly Spark methods are detected


ccordoba12:
@svanhooser, do you plan to finish this one?

delta003:
@svanhooser Going to close this, I think if you decide to finish this, it should go in our internal implementation which is more Spark / Transforms aware LS

PR #757, opened 2020-02-29
Fix flake8 io deadlock

### LINKED ISSUE #751: flake8 plugin using fragile stderr.read() call

rdugan:
Second attempt at fix to address https://github.com/palantir/python-language-server/issues/751, including test updates.  

Fixes #751.


ccordoba12 (approved):
Thanks for your help @rdugan! And sorry for the delay in merging your work.

PR #762, opened 2020-03-17
Escape Jedi path completions

### LINKED ISSUE #5519: Add path autocompletion to the Editor
## Description of your problem

**What steps will reproduce the problem?**

1. Path autocompletion, i.e. 'C:/MyFiles/<tab>' is not available in the editor window, only in the console.

**What is the expected output? What do you see instead?**

This would be a really useful enhancement from a data processing perspective. One of the features I love about Rstudio is that I can locate my files in deeply nested directories very quickly using tab-autocompletion.

andfoy:
This PR normalizes path completions returned by Jedi and escapes them in order to prevent conflicts with the code snippets grammar.

Fixes https://github.com/spyder-ide/spyder/issues/5519

### DIFF
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -1,7 +1,7 @@
 # Copyright 2017 Palantir Technologies, Inc.
 import logging
-import parso
 import os.path as osp
+import parso
 from pyls import hookimpl, lsp, _utils
 
 log = logging.getLogger(__name__)
@@ -137,7 +137,6 @@ def _format_completion(d, include_params=True):
         path = path.replace('\\', '\\\\')
         path = path.replace('/', '\\/')
         completion['insertText'] = path
-
 
     if (include_params and hasattr(d, 'params') and d.params and
             not is_exception_class(d.name)):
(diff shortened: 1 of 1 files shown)


pyls/plugins/jedi_completion.py
```diff
@@ -130,6 +132,12 @@ def _format_completion(d, include_params=True):
         'insertText': d.name
     }
 
+    if d.type == 'path':
+        path = osp.normpath(d.name)
+        path = path.replace('\\', '\\\\')
+        path = path.replace('/', '\\/')
```
ccordoba12, line 138: Do these replacements change the way you write paths on Windows and Posix systems in the Spyder editor?

andfoy: They just escape forward and backward slashes in order to prevent conflicts with the LSP snippets grammar when inserting the text. We're escaping them back in Spyder as well.

andfoy: Also, paths are normalized in order to preserve backward slashes in Windows and forward ones on POSIX

PR #780, opened 2020-04-13
Add missing test for rope_rename.py

benmezger:
This adds a missing test to `plugins/rope_rename.py` improving coverage.


test/plugins/test_rope_rename.py
```diff
@@ -0,0 +1,49 @@
...
+
+class Test2(Test1):
+    pass
+"""
+
+
+@pytest.fixture
+def tmp_workspace(workspace):
+    def create_file(name, content):
+        fn = os.path.join(workspace.root_path, name)
```
ccordoba12, line 20: Where is the `root_path` of this workspace?

benmezger: Hey @ccordoba12 sorry for the late reply, I didn't get notified about this comment.
`root_path` is currently set to `/private/var/folders/25/cy4pgsqn2393lvfmh420gf7h0000gn/T/pytest-of-benmezger/pytest-0/test_rope_rename0` on my machine.

benmezger: This is actually copied from: https://github.com/palantir/python-language-server/blob/2796154937df0d6552572c6704ff9887e2aa7b33/test/plugins/test_references.py#L26

ccordoba12: Ok, thanks for the clarification :+1:

ccordoba12 (approved):
Thanks @benmezger!

PR #784, opened 2020-04-16
Require pyflakes less than 2.2.0 for compatibility with flake8

### LINKED ISSUE #782: Conflict between required pyflakes version
Publication of pyflakes 2.2.0 version revealed a conflict in dependencies between python-language-server (pyflakes>=1.6.0) and flake8(pyflakes<2.2.0,>-2.1.0). 

I think the best solution would be to remove dependencies in python-language-server and rely only on flake8 definition.

leszekpi:
According to @ccordoba12  suggestion, I added <2.2.0 to pyflakes dependency.

Fixes #782


ccordoba12 (approved):
Thanks for your help @leszekpi!

PR #798, opened 2020-05-05
PR: Wrap Document classes inside an Agent to handle concurrent writes and reads

### LINKED ISSUE #793: Application of incremental text changes is not atomic?
In SublimeText we have this "format on save" feature. When it is enabled, and when the document is saved by the user, a textDocument/didChange is notified, and immediately after a textDocument/formatting is requested.

It can sometimes happen that pyls returns text edits that destroy / chop up my python file.

I'm looking at the code in pyls and it looks like a thread pool of some sort is used to handle requests and notifications. The textDocument/didChange must be atomic in this case, i.e. other requests/notifications must wait for the handler for textDocument/didChange to finish. I'm unsure whether I'm looking in the right corner of the codebase.

andfoy:
Fixes #793

### DIFF
--- a/pyls/plugins/rope_completion.py
+++ b/pyls/plugins/rope_completion.py
@@ -27,8 +27,8 @@ def pyls_completions(config, workspace, document, position):
 
     offset = document.offset_at_position(position)
     rope_config = config.settings(document_path=document.path).get('rope', {})
-    rope_project = workspace._rope_project_builder(rope_config)
-    document_rope = document._rope_resource(rope_config)
+    rope_project = workspace.rope_project_builder(rope_config)
+    document_rope = document.rope_resource(rope_config)
 
     try:
         definitions = code_assist(rope_project, document.source, offset, document_rope, maxfixes=3)
--- a/pyls/plugins/rope_rename.py
+++ b/pyls/plugins/rope_rename.py
@@ -13,7 +13,7 @@ log = logging.getLogger(__name__)
 @hookimpl
 def pyls_rename(config, workspace, document, position, new_name):
     rope_config = config.settings(document_path=document.path).get('rope', {})
-    rope_project = workspace._rope_project_builder(rope_config)
+    rope_project = workspace.rope_project_builder(rope_config)
 
     rename = Rename(
         rope_project,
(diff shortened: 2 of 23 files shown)


rwols:
I don't know if this is an improvement. You seem to have added a busy-while-loop. Study [condition objects](https://docs.python.org/3/library/threading.html#condition-objects) to remove that busy-while-loop.

That said, why not put access to documents, and mutation of documents, under a simple `threading.Lock`? clangd applies this idea and it works fine:

https://github.com/llvm/llvm-project/blob/master/clang-tools-extra/clangd/DraftStore.h
https://github.com/llvm/llvm-project/blob/master/clang-tools-extra/clangd/DraftStore.cpp

andfoy:
@rwols We didn't took Locks on this one, as handling them may lead to deadlocks or livelocks. The implementation here is equivalent to actors and processes, as they are implemented on Erlang. We use a queue as a mailbox to send a message and then wait for the response, this way, we ensure that no process gets  a deadlock

rwols:
If you ensure a lock is only used from within the methods of a class like a DraftStore, and you make sure that class doesn't call its own methods, then I don't know how one would cause a deadlock.

andfoy:
Actually, acquiring a lock is equivalent in some sort of sense to wait for a message to arrive. Also, by using Actors, we don't have shared state anymore, this any other process cannot modify values directly without passing through the actor. Coming from the Erlang world, it seems more reasonable to use Actors and messages for handling concurrency, rather than locks or semaphores

rwols:
> by using Actors, we don't have shared state anymore, this any other process cannot modify values directly without passing through the actor.

All I see is an extra `threading.Thread` for no real reason, and a busy while loop.
The `queue.Queue` makes use a `threading.Condition` and `threading.Lock`, so this seems like using a lock with extra unnecessary steps.

ccordoba12:
@rwols, if you disagree with @andfoy and can provide the implementation you'd like to see, you're welcome to open a new PR about it. Otherwise we'll go with his.

rwols:
I can only give advice. I don't have the time to delve into more open-source codebases. I fear that the unnecessary complexity will not make pyls easier to maintain. But you are of course free to merge this.

ccordoba12:
@andfoy, please solve the merge conflicts in this PR.

PR #781, opened 2020-04-16
Update Jedi calls for its 0.17.0+ API

bnavigator:
As discussed in https://github.com/palantir/python-language-server/issues/744#issuecomment-614338490, fixes #744, including most recent Jedi 0.17.0

~This is a combination of #775 and dalthviz#1~ Edit: removed the jedi master branch pulls in the CIs

PR #801, opened 2020-05-09
Add support for renaming with Jedi

### LINKED ISSUE #530: Make default renaming work with Jedi
Could the rename functionality be defaulted (if rope is not installed) to jedi's renaming function since it is a base dependency? The readme doesn't list it for jedi.

fsouza:
Leverages jedi's support for refactoring to implement rename.

This requires upgrading jedi. ~I can split this PR in two if preferred (first upgrade jedi, then add rename).~ Nevermind I see that #781 exists, once that's merged I can rebase.

**Note:** Jedi refactoring requires Python 3.6+, so I added 3.6 to CI. I didn't remove Python 3.5, but I believe at this point it's probably safe to do so. Thoughts? Also let me know if you'd rather go all the way to Python 3.8.

Closes #530.


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @fsouza! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

fsouza:
Oh I see that the upgrade to jedi 0.17.0 is being handled in #781. I can rebase my changes once that's merged.

ccordoba12:
> I didn't remove Python 3.5, but I believe at this point it's probably safe to do so.

I think so too. Please use 3.6 instead of 3.5 in our CIs.

test/plugins/test_jedi_rename.py
```diff
@@ -0,0 +1,56 @@
...
+DOC = '''class Test1():
+    pass
+
+class Test2(Test1):
+    pass
+'''
+
+
+@pytest.fixture
+def tmp_workspace(workspace):
```
ccordoba12: Did you copy this fixture from another file? If that's the case, please move it to `test/fixtures.py` and remove it from all files that declare it.

test/plugins/test_jedi_rename.py
```diff
@@ -0,0 +1,56 @@
...
+
+    create_file(DOC_NAME, DOC)
+    return workspace
+
+
+@pytest.mark.skipif(LT_PY36, reason='Jedi refactoring isnt supported on Python 2.x/3.5')
+def test_jedi_rename(tmp_workspace, config):  # pylint: disable=redefined-outer-name
+    position = {'line': 0, 'character': 6}
+    DOC_URI = uris.from_fs_path(os.path.join(tmp_workspace.root_path, DOC_NAME))
+    doc = Document(DOC_URI, MockWorkspace())
```
ccordoba12: Why don't you use `tmp_worksapace` here instead of `MockWorkspace`?

test/plugins/test_jedi_rename.py
```diff
@@ -0,0 +1,56 @@
...
+            f.write(content)
+        workspace.put_document(uris.from_fs_path(fn), content)
+
+    create_file(DOC_NAME, DOC)
+    return workspace
+
+
+@pytest.mark.skipif(LT_PY36, reason='Jedi refactoring isnt supported on Python 2.x/3.5')
+def test_jedi_rename(tmp_workspace, config):  # pylint: disable=redefined-outer-name
+    position = {'line': 0, 'character': 6}
```
ccordoba12, line 29: Please add a comment above this line to mention at what token this is placed on `DOC`.

pyls/plugins/jedi_rename.py
```diff
@@ -0,0 +1,44 @@
...
+
+log = logging.getLogger(__name__)
+
+
+@hookimpl
+def pyls_rename(config, workspace, document, position, new_name):  # pylint: disable=unused-argument
+    log.debug('Executing rename of %s to %s', document.word_at_position(position), new_name)
+    kwargs = _utils.position_to_jedi_linecolumn(document, position)
+    kwargs['new_name'] = new_name
+    refactoring = document.jedi_script().rename(**kwargs)
```
ccordoba12: What does Jedi do when `rename` is called for Python 2.7 or 3.5?

ccordoba12 (commented):
Thanks for this @fsouza! I left some small comments for you (please try to address them asap, so we can include your work in our next version).

pyls/plugins/jedi_rename.py
fsouza: It raises ``NotImplementedError``. Should we wrap that exception into a nicer one?

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/jedi/api/__init__.py", line 60, in wrapper
    "No support for refactorings/search on Python 2/3.5"
NotImplementedError: No support for refactorings/search on Python 2/3.5
```

test/plugins/test_jedi_rename.py
fsouza: yeah it was copy-pasta. will improve the test file

fsouza:
Hi @ccordoba12, thanks for the feedback. I'll be able to address it later today or tomorrow morning (EDT). No need to hold the release if that doesn't work for you.

Thank you! :D

test/plugins/test_jedi_rename.py
fsouza: We can't move it because it depends on variables declared on this file. Should we create a helper function in ``text/fixtures.py`` (something like `make_temp_workspace(list_of_files)`), then update all `tmp_workspace` declarations to invoke that function?

pyls/plugins/jedi_rename.py
goanpeca: Yes please, let's catch the error.

ccordoba12: I agree with @goanpeca, Please catch the error and make this method return an empty dict with the right content so it can't processed by any client.

test/plugins/test_jedi_rename.py
ccordoba12: That would really nice, thanks!

ccordoba12:
>  I'll be able to address it later today or tomorrow morning (EDT). No need to hold the release if that doesn't work for you.

If you can do it between today and tomorrow is fine. We still have to review and merge PR #754, which we plan to do tomorrow.

pyls/plugins/jedi_rename.py
fsouza: What do we want to tell clients? I think this should return an error and pyls_jsonrpc already has logic to convert exceptions into rpc errors that clients should interpret as documented on the protocol. Returning an empty dict means the server will fail silently and clients will have no idea of why, if we keep the exception clients will see an error that says "No support for refactorings/search on Python 2/3.5".

We could wrap that message to say "no support for renaming on Python 2/3.5", but we should still raise an exception.

fsouza: Pushed a better error message. Let me know what you think. Here's an example with coc.nvim:

image

fsouza:
@ccordoba12 I think I got it all. Thanks for all the feedback!

pyls/plugins/jedi_rename.py
ccordoba12: I think that's quite good, thanks!

pyls/plugins/jedi_rename.py
```diff
@@ -0,0 +1,48 @@
...
+
+@hookimpl
+def pyls_rename(config, workspace, document, position, new_name):  # pylint: disable=unused-argument
+    log.debug('Executing rename of %s to %s', document.word_at_position(position), new_name)
+    kwargs = _utils.position_to_jedi_linecolumn(document, position)
+    kwargs['new_name'] = new_name
+    try:
+        refactoring = document.jedi_script().rename(**kwargs)
+    except NotImplementedError:
+        raise Exception('No support for renaming in Python 2/3.5 with jedi. '
```
ccordoba12:
```suggestion
        raise Exception('No support for renaming in Python 2/3.5 with Jedi. '
```

ccordoba12 (commented):
Thanks @fsouza for addressing our review so quickly! I just left a minor suggestion, otherwise looks good.

ccordoba12:
Don't worry about the failures in our tests. They are due to a new flake8 version.

ccordoba12 (approved):
Thanks @fsouza!

PR #807, opened 2020-05-12
Add support for autopep8 aggressive option from config file

andrewzigerelli:
This is so the user can specify aggressive option in pycodestyle config file.
e.g. in ~/.config/pycodestyle
```
[pycodestyle]
aggressive = 2
```
This seems to work. E.g.
No aggressive option leave this code as is:
```python
if x == None:
```
With above pycodestyle and pull request, it makes this change:
```python
if x is None:
```

Anyone who reads this: higher aggressive options than 0 should be used with care, as x is None does not always mean the same as x == None.

### DIFF
--- a/pyls/config/pycodestyle_conf.py
+++ b/pyls/config/pycodestyle_conf.py
@@ -15,6 +15,7 @@ OPTIONS = [
     ('ignore', 'plugins.pycodestyle.ignore', list),
     ('max-line-length', 'plugins.pycodestyle.maxLineLength', int),
     ('select', 'plugins.pycodestyle.select', list),
+    ('aggressive', 'plugins.pycodestyle.aggressive', int),
 ]
 
 
--- a/pyls/plugins/autopep8_format.py
+++ b/pyls/plugins/autopep8_format.py
@@ -57,6 +57,7 @@ def _autopep8_config(config):
         'ignore': settings.get('ignore'),
         'max_line_length': settings.get('maxLineLength'),
         'select': settings.get('select'),
+        'aggressive': settings.get('aggressive'),
     }
 
     # Filter out null options


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @andrewzigerelli! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

goanpeca:
Thanks for working on this @andrewzigerelli 

Could you add a small test for this ?

Using the x == None example should be enough in the test

ccordoba12:
@andrewzigerelli, please also sign the CLA agreement so we can merge your PR (we can't do it without it).

andrewzigerelli:
Does the current ci test pycodestlye config files? I'm not familiar with ci tests

goanpeca:
>Does the current ci test pycodestlye config files?
Yes

>I'm not familiar with ci tests

Take a look at this file https://github.com/palantir/python-language-server/blob/develop/test/plugins/test_pycodestyle_lint.py

andrewzigerelli:
@goanpeca 
I wrote a small test and do 
```
pytest test/plugins/test_pycodestyle_lint.py -s >2 log.out
```
This reads from my ~/.config/pycodestyle and is fine, but I don't see how this will pass ci validation.
Do I have to install circleci locally to see [setup.cfg](https://github.com/palantir/python-language-server/blob/develop/setup.cfg) be preferred?

Further, [test_pycodestyle_lint.py](https://github.com/palantir/python-language-server/blob/develop/test/plugins/test_pycodestyle_lint.py) doesn't even seem to test the options defined in setup.cfg, unless I'm in the wrong file. I don't see any code which tests max-line-length = 120, or the ignore options..

goanpeca:
Actually I think is fine to leave without a test for now :-)

goanpeca (approved):

PR #803, opened 2020-05-11
Move CI to Github Actions

goanpeca:
Add win, mac and linux testing for Python 2.7, 3.6, 3.7 and 3.8

PR #809, opened 2020-05-12
Refactor: Remove Python 2 support from code-base

benmezger:
The PR is a work-in-progress which removes Python 2 support from PYLS.  Config and _utils.py. At the moment, have been typed and refactored a little.
While working on this, I encountered a few questions that would be nice to discuss. Should we type all functions/methods? Or should we type only API/abstraction related code? Nevertheless, if we stick with typing, I figured it would be interesting to have `mypy` running on CI to make sure code is correctly typed. 

Currently there is quite a lot of logic within some dunder inits, so I moved some of them to a method of their own, separating the imports logic from the initialization itself. 

This is running on Python 3.7.0. I had some trouble running on Python 3.8.x due to pickling issues when running the completion tests, so I will have to investigate this a little further.

Related issue: #737


goanpeca:
Hi @benmezger, thanks for working on this.

>Nevertheless, if we stick with typing, I figured it would be interesting to have mypy running on CI to make sure code is correctly typed.

Sure that would be nice moving forward, as long as we can do typing incrementally. Some methods might need to be refactored in order for them to make sense with typing. So as long as not much refactoring is needed, adding typing would be :+1: (on my side). If too much refactoring is needed on a given func/method, I would leave that for another PR.

For the Py38 issue you might need to do something like this -> https://github.com/palantir/python-language-server/pull/803/files#diff-5de261d12a7ae8f098419c22b40ce213R34

You will need to remove py27 testing from CI files (circle ci and appveyor) to get all tests running?

benmezger:
> For the Py38 issue you might need to do something like this -> https://github.com/palantir/python-language-server/pull/803/files#diff-5de261d12a7ae8f098419c22b40ce213R34

Thank you for this link, I will try this out later this evening. I usually type most abstract classes or any API class around the code-base, but that something per-project, so I will type/refactor what I find relevant for now and we can discuss further in this PR, is that ok?

About the CI, I will remove it on the next commit.

goanpeca:
>what I find relevant for now and we can discuss further in this PR, is that ok?

Sure thing 👍

setup.py
```diff
@@ -101,4 +101,5 @@
             'yapf = pyls.plugins.yapf_format'
         ]
     },
+    python_requires='>=3,<=3.7.0'
```
benmezger, line 104: This may require some further discussion, as I am not sure if allowing a pip install of a non compatible Python version is very smart

benmezger:
@ccordoba12  I have a few related branches which took-off from this branch, should I make a PR for each branch or merge it over here? Changes are related to code-base change, old class style removal etc.

ccordoba12:
> Changes are related to code-base change, old class style removal etc.

I agree with removing old class style declarations here. If you want to perform further refactorings (style or code wise), please leave them for other PRs.

PR #811, opened 2020-05-14
Fix completion with one arg and add optional class objects to completion list

mpanarin:
1. Fixes the issue of 1 arg completion snippet.
Before the `$0` would be inserted literally, as insertTextFormat was missing.

2. Allows adding class objects to completion. 
When you want to call a classmethod or staticmethod, class name is completed with snippet as if we are instantiating it. This is super annoying, especially if init has a lot of args.

### DIFF
--- a/pyls/lsp.py
+++ b/pyls/lsp.py
@@ -24,6 +24,13 @@ class CompletionItemKind(object):
     Color = 16
     File = 17
     Reference = 18
+    Folder = 19
+    EnumMember = 20
+    Constant = 21
+    Struct = 22
+    Event = 23
+    Operator = 24
+    TypeParameter = 25
 
 
 class DocumentHighlightKind(object):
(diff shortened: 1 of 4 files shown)


goanpeca:
Awesome @mpanarin !

- Did a quick pass and looks good. Just one question should the `should_include_class_objects` option be `False` by default?
- And does jedi offers such configuration or is something we would be doing just on the PyLS side?

Thanks!

mpanarin:
@goanpeca 
1. I updated that part. I think the only people who would be annoyed by this features are the ones, who have snippets disabled. So I added a check for that and made `should_include_class_objects` enabled by default
2. No, I don't think so. We submit an issue or PR to jedi, but is it worth it?

goanpeca:
1 👍🏼
2 No need, just wondering about that.

Thanks!

benmezger:
Thanks for this. Been having issues with the annoying arg completion, just tried out your fork and seems to be working fine over here.

mpanarin:
@goanpeca @ccordoba12 
As this one is ready, can it be merged?
Thanks!

goanpeca (approved):
Thanks for the work @mpanarin !

It works as expected on my local tests @ccordoba12

PR #814, opened 2020-05-25
Fix an issue when re.match returns None

mpanarin:
`re.match` can return `None`, so we can't just simply call `.object` on the result.


ccordoba12 (approved):
Looks good to me, thanks @mpanarin!

PR #818, opened 2020-06-12
Do not create documents from a textDocument/rename

### LINKED ISSUE #816: The returned VersionedTextDocumentIdentifiers from textDocument/rename have a too high document version
When doing a textDocument/rename, pyls will nicely suggest to open non-tracked files to open and apply the changes to.

However, the `"version"` of the files to-be-opened have their version number too high.

Here's a part of the response payload:
```js
        {
            "textDocument": {
                "uri": "file:///home/raoul/.config/sublime-text-3/Packages/LSP/plugin/references.py",
                "version": 9
            },
            "edits": [
                {
                    "range": {
                        "start": {
                            "line": 0,
                            "character": 0
                        },
                        "end": {
                            "line": 182,
                            "character": 0
                        }
                    },
                    "newText": "<snip>"
                }
            ]
        },
```

The file references.py was not yet open in the editor.
In case the document is not yet tracked via a textDocument/didOpen, I suggest to put `"version": null` here as that's allowed per spec.

Right now, the `"version": 9` came from the file that ran the textDocument/rename request. It had a version number of 9 and somehow that number was used for all the files that weren't yet tracked with textDocument/didOpen.

For a newly-opened file, our internal version counter starts at `0`. If the version of the edit doesn't match we don't apply the edit, as can be seen from the console:

```
LSP: ignoring edit due to non-matching document version
```

python-language-server version: 0.33.0

rwols:
During a textDocument/rename, it is possible that Jedi proposes changes in files
that have not yet been opened with textDocument/didOpen. In that case, we must
put "version": None as the document version.

Note that rope does not have this behavior.

closes #816

### DIFF
--- a/pyls/plugins/jedi_rename.py
+++ b/pyls/plugins/jedi_rename.py
@@ -17,30 +17,29 @@ def pyls_rename(config, workspace, document, position, new_name):  # pylint: dis
         raise Exception('No support for renaming in Python 2/3.5 with Jedi. '
                         'Consider using the rope_rename plugin instead')
     log.debug('Finished rename: %s', refactoring.get_diff())
-
-    return {
-        'documentChanges': [
-            {
-                'textDocument': {
-                    'uri': uris.uri_with(document.uri, path=file_path),
-                    'version': workspace.get_document(document.uri).version,
-                },
-                'edits': [
-                    {
-                        'range': {
-                            'start': {'line': 0, 'character': 0},
-                            'end': {
-                                'line': _num_lines(changed_file.get_new_code()),
-                                'character': 0,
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @rwols! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12 (approved):
Looks good to me, thanks @rwols!

rwols:
Thanks!

PR #820, opened 2020-06-19
Do not start shutdown sequence on TCP when not checking parent process

### LINKED ISSUE #12598: Connecting to external LSP servers is broken
Since the introduction of the code completion manager some months ago, it's not possible to connect to external LSP servers.

andfoy:
Fixes https://github.com/spyder-ide/spyder/issues/12598


ccordoba12 (approved):
Looks good to me, thanks @andfoy!

PR #821, opened 2020-06-20
Add configurable flake8 executable

ruhulio:
The context for this change:
- I'm using virtural envs with different versions of `Python` so each of them has their own install of `pyls` and dependencies.
- I'm using `Sublime Text` with [`Sublime LSP`](https://github.com/sublimelsp/LSP) configured to use the `pyls` from the project's virtual env.
    <details>
    <summary>Example configuration</summary>

    ```json
    {
      "settings": {
        "LSP": {
          "pyls": {
            "enabled": true,
            "command": [
              "/virtualenvs/important-project/bin/pyls"
            ],
            "settings": {
              "pyls": {
                "plugins": {
                  "flake8": {
                    "enabled": true,
                    "config": "/work/code/important-project/.flake8"
                  }
                }
              }
            }
          }
        }
      }
    }
    ```
    </details>

Within that context, the `flake8_lint` plugin tries to start `flake8` via a process; either directly or as a module. In both cases, this happens outside of the virtual env so it does not pick up the correct `flake8` executable or module.

This change allows for adding an optional explicit `executable` configuration for `flake8` that can point directly to the virtual env's installed version.

This change should be backwards compatible as it defaults to the previous behavior when there is no `executable` provided in the configuration.

The other plugin's do not have a similar issue as they do not use a separate process.

### DIFF
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -39,22 +39,24 @@ def pyls_lint(config, document):
         log.debug("using flake8 with config: %s", opts['config'])
 
     # Call the flake8 utility then parse diagnostics from stdout
+    flake8_executable = settings.get('executable', 'flake8')
+
     args = build_args(opts, document.path)
-    output = run_flake8(args)
+    output = run_flake8(flake8_executable, args)
     return parse_stdout(document, output)
 
 
-def run_flake8(args):
+def run_flake8(flake8_executable, args):
     """Run flake8 with the provided arguments, logs errors
     from stderr if any.
     """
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @ruhulio! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12 (approved):
Thanks for your contribution @ruhulio!

PR #827, opened 2020-06-27
Read pycodestyle and flake8 configurations per workspace

### LINKED ISSUE #819: Workspace folders and linter settings
If I have two workspace folders `a` and `b`, and there's `a/tox.ini` and `b/tox.ini`, then pycodestyle seems to only use the linter settings from whatever file was opened first in one of the workspace folders.

So I have line length set to 80 in `a/tox.ini`, and line length set to 120 in `b/tox.ini`, and if I open a file in `a`, then all files opened in `b` will get linter warnings about line lengths not respecting 80 columns max.

andfoy:
This also gives more priority to the user config over the client one, so the last doesn't overshadow the first.

Fixes https://github.com/palantir/python-language-server/issues/819

### DIFF
--- a/pyls/config/config.py
+++ b/pyls/config/config.py
@@ -106,6 +106,13 @@ class Config(object):
         settings = {}
         sources = self._settings.get('configurationSources', DEFAULT_CONFIG_SOURCES)
 
+        # Plugin configuration
+        settings = _utils.merge_dicts(settings, self._plugin_settings)
+
+        # LSP configuration
+        settings = _utils.merge_dicts(settings, self._settings)
+
+        # User configuration
         for source_name in reversed(sources):
             source = self._config_sources.get(source_name)
             if not source:
@@ -113,14 +120,8 @@ class Config(object):
             source_conf = source.user_config()
             log.debug("Got user config from %s: %s", source.__class__.__name__, source_conf)
             settings = _utils.merge_dicts(settings, source_conf)
-        log.debug("With user configuration: %s", settings)
 
-        settings = _utils.merge_dicts(settings, self._plugin_settings)
-        log.debug("With plugin configuration: %s", settings)
-
-        settings = _utils.merge_dicts(settings, self._settings)
-        log.debug("With lsp configuration: %s", settings)
-
+        # Project configuration
         for source_name in reversed(sources):
             source = self._config_sources.get(source_name)
             if not source:
@@ -128,7 +129,8 @@ class Config(object):
             source_conf = source.project_config(document_path or self._root_path)
             log.debug("Got project config from %s: %s", source.__class__.__name__, source_conf)
             settings = _utils.merge_dicts(settings, source_conf)
-        log.debug("With project configuration: %s", settings)
+
+        log.debug("With configuration: %s", settings)
 
         return settings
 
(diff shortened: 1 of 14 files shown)


mrclary:
Unfortunately, this PR does not fix spyder-ide/spyder#13110. 
After pulling this into my Spyder subrepo (installed during bootstrap) the same behavior I described [here](https://github.com/spyder-ide/spyder/issues/13110#issuecomment-650418791) is exhibited. `watched_folders` doesn't seem to do anything.

mrclary:
@ccordoba12 , your recent changes have not fixed the issue.
The log entries "Reading config from " and "Got user config from PyCodeStyleConfig" confirm that configurations files are found and configurations loaded from `~/.config/pycodestyle` and `path/to/project/pycodestyle.cfg`, but are still ineffective. If a configuration file is placed in `~/.spyder-py3-dev/lsp-root-path/`, it is effective.

ccordoba12:
> @ccordoba12 , your recent changes have not fixed the issue

They have. I'll open a PR in the Spyder repo with a test for this, so you can see that that's the case.

mrclary:
Perhaps I'm doing something wrong, then.
I've pulled these changes into my fork of python-language-server, then used git-subrepo to update it in the subrepo of my local spyder repo, then run from bootstrap, which installs and uses the python-language-server from the spyder subrepo.
I've reset Spyder configuration settings, then enable code style linting.
If I place a pycodestyle.cfg in a project and open that project, the lsp logs indicate that the file is found and the configuration is loaded, but the linting in the editor is wrong. If I place that same config file in ~/spyder-py3-dev/lsp-root-path, then it works.

ccordoba12:
Sorry, you were right. There's a single case where this works right now: when Spyder starts with an open project. Otherwise, after switching projects Spyder needs to tell Jedi to extend its `sys.path` with the project's directory. And that overwrites the pycodestyle config read from a file with our own (because we're sending our entire config, there's no way to send just a fraction of it).

mrclary:
@ccordoba12 , this works well on my end, now.
I don't want to be too fastidious, but for an open project pycodestyle.cfg is monitored for changes (i.e. "Handling notification from client {'method': 'workspace/didChangeWatchedFiles' ...") while outside of a project ~/.config/pycodestyle is not monitored and Spyder must be restarted to load any changes. Is it possible to add the user config file to watched files when outside of a project?

ccordoba12:
> this works well on my end, now.

Great to know!

> outside of a project ~/.config/pycodestyle is not monitored and Spyder must be restarted to load any changes

That's strange because that file is read every time a linting request is processed. I'll take a look at it.

ccordoba12:
> outside of a project ~/.config/pycodestyle is not monitored and Spyder must be restarted to load any changes

Sorry but I took a quick look at this and I couldn't find an easy way to add that functionality. Since the error you reported is already fixed, I think that's enough for now.

ccordoba12:
Of course, you're welcome to open a PR to solve that problem. I agree it'd be nice to have, but it's not critical.

mrclary:
> Of course, you're welcome to open a PR to solve that problem. I agree it'd be nice to have, but it's not critical.

Agreed. Thanks a lot for taking care of this.

PR #830, opened 2020-07-02
Make flake8 plugin use stdin

### LINKED ISSUE #743: Flake8 plugin does not check content of unsaved buffer
It seems like the way the plugin is written it uses the document.path in the flake8 argument to check for errors:

https://github.com/palantir/python-language-server/blob/f6df42cae186e5389e0332e9b8406bb24bd84d3e/pyls/plugins/flake8_lint.py#L42-L45

Where pycodestyle would actually use the document.lines instead:

https://github.com/palantir/python-language-server/blob/f6df42cae186e5389e0332e9b8406bb24bd84d3e/pyls/plugins/pycodestyle_lint.py#L33-L35

This seems wrong as the flake8 check would ignore the update to the current buffer and always check what is written to disk.  From the editor, you basically see all the signs/error lines to be incorrect.

For example, I'm using vim and use vim-lsp-settings, I have a configuration like:

```
    let g:lsp_settings['pyls'] = {
    \    'workspace_config': {'pyls': {'plugins': {
    \      'flake8': {'enabled': v:true,},
    \      'pycodestyle': {'enabled': v:false,},
    \      'pyflakes': {'enabled': v:false},
    \      'mccabe': {'enabled': v:false},
    \    }}}
    \  }
```

Say if you have a test.py file with this:
```
print(y)
```
Save the file and edit again, you'd see the line is flagged as error.

However, if you insert the line above it and add another print:

```
print(x)
print(y)
```
The result is that the error would still only appear on line 1 and NOT both lines since flake8 is run on the file saved on disk.  It does not pipe the buffer to it instead.

Flake8 can read from stdin so it should be possible to pipe the lines to the checker.  For instance:

```
cat test.py | flake8 -
```

### LINKED ISSUE #825: test "test_flake8_no_checked_file" failing on python 3.9
In trying to build pyls on Fedora rawhide (to be 33), I get the test failure below - 

```_________________________ test_flake8_no_checked_file __________________________
config = <pyls.config.config.Config object at 0xec7c6418>
    def test_flake8_no_checked_file(config):
        # A bad uri or a non-saved file may cause the flake8 linter to do nothing.
        # In this situtation, the linter will return an empty list.
    
        doc = Document('', DOC)
        diags = flake8_lint.pyls_lint(config, doc)
>       assert diags == []
E       assert [{'code': 'E9...ity': 2, ...}] == []
E         Left contains one more item: {'code': 'E902', 'message': "FileNotFoundError: [Errno 2] No such file or directory: ''", 'range': {'end': {'character': 10, 'line': -1}, 'start': {'character': 0, 'line': -1}}, 'severity': 2, ...}
E         Full diff:
E           [
E         -  ,
E         +  {'code': 'E902',
E         +   'message': "FileNotFoundError: [Errno 2] No such file or directory: ''",
E         +   'range': {'end': {'character': 10,...
E         
E         ...Full output truncated (7 lines hidden), use '-vv' to show
test/plugins/test_flake8_lint.py:37: AssertionError
```

Versions of dependencies are here - https://kojipkgs.fedoraproject.org//work/tasks/1161/46111161/root.log

I can provide other info as needed and/or test patches.

qbedard:
This resolves #743. It may also resolve #825, but I have not tested to be sure.

### DIFF
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -43,12 +43,12 @@ def pyls_lint(workspace, document):
     # Call the flake8 utility then parse diagnostics from stdout
     flake8_executable = settings.get('executable', 'flake8')
 
-    args = build_args(opts, document.path)
-    output = run_flake8(flake8_executable, args)
+    args = build_args(opts)
+    output = run_flake8(flake8_executable, args, document)
     return parse_stdout(document, output)
 
 
-def run_flake8(flake8_executable, args):
+def run_flake8(flake8_executable, args, document):
     """Run flake8 with the provided arguments, logs errors
     from stderr if any.
     """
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @timbedard! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12 (approved):
Thanks for your help @timbedard! Looks good to me.

qbedard:
Happy to do what I can. 😁

PR #418, opened 2018-09-09
Pass document path to plugin settings

gatesn:
Fixes #413

PR #845, opened 2020-08-22
Specify relevant document path when retrieving some plugin settings

mnauw:
... particularly those that are extracted from configuration files
relative to document location

See also PR #418


ccordoba12 (commented):
Thanks for your contribution @mnauw! This looks good to me, but please apply your changes also to the Mccabe and Pydocstyle linters.

mnauw:
As suggested, the changes have also been applied to the Mccabe and Pydocstyle linters (and rebased on current `develop`).

ccordoba12 (approved):
Looks good to me, thanks @mnauw!

PR #846, opened 2020-08-25
Correct method and attribute detection on document/symbols call

andfoy:
This PR corrects the types of methods and attributes, which were marked as functions and variables by Jedi, respectively


ccordoba12 (approved):
Thanks @andfoy, great improvement!

PR #851, opened 2020-09-05
PR: Fix compatibility with Jedi master

Mic92:
- Newer versions of parso/jedi uses Path instead of str
- by converting to str pyls is compatibility with both old and new versions

### DIFF
--- a/pyls/plugins/definition.py
+++ b/pyls/plugins/definition.py
@@ -16,7 +16,7 @@ def pyls_definitions(config, document, position):
 
     return [
         {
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @Mic92! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

delta003 (approved):

PR #842, opened 2020-08-12
Update ujson dependency

steff456:
This PR updated the ujson dependency.

Fixes #857 
Fixes #802

PR #859, opened 2020-09-10
Pinned `python-jsonrpc-server` to < 0.4

zerocewl:
Pinned `python-jsonrpc-server` to < 0.4 to avoid ujson error

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ setup(
         'future>=0.14.0; python_version<"3"',
         'backports.functools_lru_cache; python_version<"3.2"',
         'jedi>=0.17.0,<0.18.0',
-        'python-jsonrpc-server>=0.3.2',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -36,7 +36,7 @@
         'future>=0.14.0; python_version<"3"',
         'backports.functools_lru_cache; python_version<"3.2"',
         'jedi>=0.17.0,<0.18.0',
-        'python-jsonrpc-server>=0.3.2',
+        'python-jsonrpc-server>=0.3.2,<0.4',
```
phijor, line 39: Why not

```suggestion
        'python-jsonrpc-server~=0.3.2',
```

? I think that would show the intent of "version `0.3.2` or any later, but compatible version".

ccordoba12:
This won't be necessary when PR #842 is merged, but thanks for your concern.

PR #866, opened 2020-09-24
Synchronize document text updates

### LINKED ISSUE #793: Application of incremental text changes is not atomic?

andfoy:
Fixes #793
Closes #798

### DIFF
--- a/pyls/workspace.py
+++ b/pyls/workspace.py
@@ -3,6 +3,8 @@ import io
 import logging
 import os
 import re
+import functools
+from threading import RLock
 
 import jedi
 
@@ -13,6 +15,14 @@ log = logging.getLogger(__name__)
 # TODO: this is not the best e.g. we capture numbers
 RE_START_WORD = re.compile('[A-Za-z_0-9]*$')
 RE_END_WORD = re.compile('^[A-Za-z_0-9]*')
+
+
+def lock(method):
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+        with self._lock:
+            return method(self, *args, **kwargs)
+    return wrapper
 
 
 class Workspace(object):
@@ -131,6 +141,7 @@ class Document(object):
         self._source = source
         self._extra_sys_path = extra_sys_path or []
         self._rope_project_builder = rope_project_builder
+        self._lock = RLock()
 
     def __str__(self):
         return str(self.uri)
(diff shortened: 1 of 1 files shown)


pyls/workspace.py
```diff
@@ -15,6 +17,14 @@
 RE_END_WORD = re.compile('^[A-Za-z_0-9]*')
 
 
+def lock(method):
```
ccordoba12, line 20: Please add a docstring to this function.

pyls/workspace.py
```diff
@@ -218,6 +232,7 @@ def word_at_position(self, position):
 
         return m_start[0] + m_end[-1]
 
+    @lock
     def jedi_names(self, all_scopes=False, definitions=True, references=False):
```
ccordoba12, line 237: Why does `jedi_names` need to be locked, but not the other Jedi operations?

ccordoba12 (commented):
Thanks @andfoy for this! Could you mention the motivation for these changes in the PR description?

Also, is it possible to add a test for this?

pyls/workspace.py
ccordoba12: Since you added the `lock` decorator to `jedi_script`, does the one above this line is still necessary?

ccordoba12 (commented):
@andfoy, is it possible to add a test for this?

andfoy:
It is not possible to guarantee a certain scenario for concurrent accesses, since it is not possible to ensure that a specific thread arrives before other

pyls/workspace.py
ccordoba12: What do you think about what I said in my comment above? Do you really need to add the `lock` decorator to `jedi_names`?

ccordoba12 (approved):
Thanks @andfoy, looks good to me.

PR #885, opened 2020-11-04
Add code in front of the diagnostic message for flake8

### LINKED ISSUE #884: It would be better to keep output from the flake8 linter to be consistent with that of pycodestyle.
Currently, the return message the flake8 linter [here](https://github.com/palantir/python-language-server/blob/develop/pyls/plugins/flake8_lint.py) does not contain the error code, while that of pycodestyle also include the error code.

It would be better to include the error code as well, something like the following:

> E303 too many blank lines (2)

It is a fairly simple fix.  Should I open a pull request for this?

jdhao:
For the issue discussed in https://github.com/palantir/python-language-server/issues/884, it would be better to also add the code in front the actual diagnostic message, to make the message style the same as pycodestyle.

Fixes #884.

### DIFF
--- a/pyls/plugins/flake8_lint.py
+++ b/pyls/plugins/flake8_lint.py
@@ -144,6 +144,8 @@ def parse_stdout(document, stdout):
         _, line, character, code, msg = parsed_line
         line = int(line) - 1
         character = int(character) - 1
+        # show also the code in message
+        msg = code + ' ' + msg
         diagnostics.append(
             {
                 'source': 'flake8',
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @jdhao! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ccordoba12:
@jdhao, thanks a lot for this! Probably our tests for the flake8 plugin are failing due to the change you introduced, so you also need to fix them in this PR.

jdhao:
> @jdhao, thanks a lot for this! Probably our tests for the flake8 plugin are failing due to the change you introduced, so you also need to fix them in this PR.

OK, it seems that I also need to update some of the test for flake8 plugin, as the msg format has changed.

jdhao:
@ccordoba12  I have fixed the bug. Please review it.

ccordoba12 (approved):
Looks good to me, thanks @jdhao!

PR #888, opened 2020-11-20
PR: Enable code folding results aggregation

andfoy:
This PR enables other pyls plugins to add their own folding results to the ones computed by the main server. While there exist the possibility of having folding start collisions, plugin creators should prevent this.


ccordoba12 (approved):
Thanks @andfoy.

PR #893, opened 2020-12-10
PR: Flatten folding regions

andfoy:
This PR applies list flattening in order to aggregate multiple code folding hook implementation results.

Required by https://github.com/spyder-ide/pyls-spyder/pull/15


ccordoba12 (approved):
Looks good to me, thanks @andfoy!

PR #905, opened 2021-01-30
Implement completionItem/resolve

### LINKED ISSUE #903: Implement completionItem/resolve
This is a proposal to implement `completionItem/resolve` in order to improve the speed of completions; the need to resolve `documentation` dynamically was confirmed by profiling in #826.

The idea is that the `textDocument/completion` sends only the properties which are cheap to compute (say `label`, `sortText`, `insertText`, etc), while `documentation`, `detail` (and any other property since LSP 3.16.0) would be only sent once the clients requests to resolve the details with the `completionItem/resolve` request. This improves the initial speed and gives the user all information - no compromise.

The lag for `completionItem/resolve` can be lower than the perception threshold (dozens of milliseconds), and the client can pre-fetch a few `completionItem/resolve` as soon as it gets the completion result (I implemented this in my client [here](https://github.com/krassowski/jupyterlab-lsp/pull/487)).

Would any of the maintainers be interested in accepting a PR that would do that? Any design suggestions?

I am opening this issue before starting any work as I see multiple PR hanging without response. Do I understand that this repository is now maintained by volunteers and not by the palantir anymore?

krassowski:
Fixes #903

- implemented new hook for resolve
- implemented resolve in jedi_completion and rope_completion
- added `eager` option to rope_completion and jedi_completion allow users using clients which cannot yet handle `completionItem/resolve` to restore to eager resolution (inside of `textDocument/completions`)
- added test for resolve and documentation in Jedi
- tested with development version of [JupyterLab-LSP](https://github.com/krassowski/jupyterlab-lsp) client; the client implementation was tested against reference JavaScript/TypeScript server and against the R-languageserver

### DIFF
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -28,6 +28,11 @@ def pyls_completions(config, workspace, document, position):
 
 
 @hookspec
+def pyls_completion_item_resolve(config, workspace, completion_item):
+    pass
+
+
+@hookspec
 def pyls_definitions(config, workspace, document, position):
     pass
 
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -27,7 +27,7 @@ def pyls_completions(config, workspace, document, position):
     pass
 
 
-@hookspec
+@hookspec(firstresult=True)
 def pyls_completion_item_resolve(config, workspace, completion_item):
     pass
 
--- a/pyls/plugins/jedi_completion.py
+++ b/pyls/plugins/jedi_completion.py
@@ -56,6 +56,8 @@ _LAST_COMPLETIONS = {}
 @hookimpl
 def pyls_completions(config, document, position):
     """Get formatted completions for current code position"""
+    # pylint: disable=too-many-locals
+    # pylint: disable=global-statement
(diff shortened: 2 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/python-language-server](https://github.com/palantir/python-language-server), @krassowski! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

krassowski:
@ccordoba12 thanks for expressing your interest! The PR is now green and ready for review.

krassowski:
Cross-referencing my other implementation in https://github.com/pappasam/jedi-language-server/pull/56 which is very similar but some details differ mostly and it might be helpful to compare those during review.

krassowski:
I believe that the CI failures after last commit are caused by a new release of a dependency used in tests (numpy?) and not by my code. It also explains why only the modern versions (Py3.7 and Py3.8) fail.

andfoy:
@krassowski, thanks for this PR, right now we lost our permissions on this repo, and we're waiting to see if it is going to be moved to a fork

PR #946, opened 2021-12-08
Add formatter options support for yapf

### LINKED ISSUE #328: Code formatting configuration
Hi! Thanks for `pyls`, it's great.

Historically, YAPF was the preferred formatter in `pyls`. This recently changed with #315 without a rationale provided (in fact the README still states the opposite). There's a few problems with this.

1. I don't think defaulting to autopep8 is a great choice. [YAPF's README](https://github.com/google/yapf) provides a good explanation as to why.

2. It seems to me like the only option to configure an alternative formatter is to uninstall YAPF and/or autopep8 from the virtualenv?

It would be great if there was a way for people to configure which one they want, ideally on a per-project basis.

masad-frost:
resolves https://github.com/palantir/python-language-server/issues/328 

This PR adds support for LSP formatting options https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#formattingOptions

It adds explicit support for `tabSize` and `insertSpaces`, the rest of the options are not configurable in yapf. I did however add support for arbitrary configurations as specified in `[key: string]: boolean | integer | string;`.

I think this PR might break plugins as the API now passes `options` to formatter hooks

### DIFF
--- a/pyls/hookspecs.py
+++ b/pyls/hookspecs.py
@@ -73,12 +73,12 @@ def pyls_folding_range(config, workspace, document):
 
 
 @hookspec(firstresult=True)
-def pyls_format_document(config, workspace, document):
+def pyls_format_document(config, workspace, document, options):
     pass
 
 
 @hookspec(firstresult=True)
-def pyls_format_range(config, workspace, document, range):
+def pyls_format_range(config, workspace, document, range, options):
     pass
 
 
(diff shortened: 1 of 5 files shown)


masad-frost:
closing in favor of https://github.com/python-lsp/python-lsp-server/pull/134 since this repo is not maintained

PR #982, opened 2026-09-16
Stop treating non-file document URIs as filesystem paths

ZayanKhan-12:
## Description

Partially addresses #199.

Editors that keep buffers in memory send URIs with no file behind them — Monaco's `inmemory://dummy.py` from the original report, and VS Code's `untitled:` and `vscode-notebook-cell:`. The server assumed every document had a file, derived a path, and then used that path for filesystem work.

The specific trigger is that `inmemory://dummy.py` puts the name in the URI **authority**, not the path, so `urlparse` gives `netloc='dummy.py', path=''`. The UNC branch in `to_fs_path` is gated on `scheme == 'file'`, so the authority was dropped and the function returned the empty string:

```
inmemory://dummy.py    -> to_fs_path ''       -> filename ''  dot_path ''
untitled:Untitled-1    -> to_fs_path 'Untitled-1'
```

Three things followed from that empty path. The one worth highlighting is the second:

1. Jedi received `path=''` rather than no path.
2. **`os.path.dirname('')` normalizes to `'.'`**, so asking for completions with `use_document_path=True` put the **server's own working directory** on Jedi's module search path. Completions in an in-memory buffer could resolve against whatever happened to sit next to the running process. Verified by capturing the `sys_path` handed to `jedi.Project`.
3. Reading a document that had not been opened yet fell through to `io.open('')`, raising `FileNotFoundError: ''` — an error naming a path the client never sent.

## Changes

`to_fs_path` keeps the authority for non-file schemes, so `inmemory://dummy.py` resolves to `/dummy.py` and the document has a stable identity for module naming and diagnostics.

**This deliberately differs from `vscode-uri`**, which drops the authority and returns the empty path. Since the module docstring points at `vscode-uri` and a reader may well diff the two, the reason is recorded in a comment next to the branch. File URIs, UNC paths and Windows drive letters are untouched:

```
file:///tmp/a.py       -> /tmp/a.py        (unchanged)
file://share/c$/a.py   -> //share/c$/a.py  (unchanged)
file:///C:/far/boo     -> c:/far/boo       (unchanged)
inmemory://dummy.py    -> /dummy.py        (was '')
```

Adds `uris.is_file_uri()` and `Document.is_file_backed` so callers can tell a path-shaped *identifier* from an actual *location*, then uses it to skip the `sys.path` extension for non-file documents and to raise a clear error instead of reading an unrelated file.

## Scope

This covers **in-memory analysis of a single document**, which is what @gatesn asked about in the thread ("purely in-memory analysis, e.g. only things passed by didOpen").

It does **not** implement what @ColinTenaguillo asked for in 2022 — importing between two in-memory documents. That needs Jedi to see unsaved buffers as importable modules, which means a custom project or importer rather than a URI fix, and I did not want to bundle it into this change. Flagging it so the issue is not closed as fully resolved on the strength of this PR.

## Tests

16 new tests across `test/test_uris.py` and `test/test_document.py`: the URI translations above, `is_file_uri` across schemes, the document properties, the `sys.path` behaviour for both document kinds, the error when an unopened non-file document is read, and a completion inside an in-memory document.

I checked the tests actually hold the line by re-breaking each fix in turn: reverting the authority handling fails 3, restoring the `sys.path` injection fails 1, removing the source guard fails 1.

## Verification

Run on Python 3.8 with `jedi 0.17.2`, matching the CI matrix.

**A clean checkout of `develop` does not produce a clean run**, so I recorded a baseline first and compared:

| | baseline | with this change |
| --- | --- | --- |
| `pytest test/` | 12 failed, 99 passed, 8 skipped | 12 failed, **115 passed**, 8 skipped |
| failing test names | — | **identical set** |
| `pycodestyle pyls test` | clean | clean |
| `pyflakes pyls test` | 1 pre-existing (`_utils.py`) | unchanged |

The 12 failures are pre-existing and track linter versions that moved on since 2020 (flake8, pydocstyle, pylint, numpy hover). None are touched by this change.

`pylint pyls test` goes from 84 messages to 92. The eight added are 6 × `redundant-u-string-prefix` and 2 × `consider-using-f-string`, both of which the **Python 2.7 support in this repo requires** and both already present in the baseline. I fixed the one genuinely new message I introduced, a `wrong-import-order` caused by `test` shadowing a stdlib name.

Two environment pins were needed to get the suite running at all, neither related to this change: `setuptools<70` (`pkg_resources` was removed) and `pylint<3` (`pylint.epylint` was removed). Both are captured in the `CLAUDE.md` added in the second commit, which is independent — happy to drop that commit if it is not wanted.


palantirtech:

Issue #834, opened 2020-07-21
Unable to run pyls with VS code


CHRC-Int:
Hello

I'm trying to get develop this against VS code but when I run `yarn install` I'm warned that

    yarn install v1.22.4
    [1/5] Validating package.json...
    warning pyls@0.0.1: The engine "vscode" appears to be invalid.
    [2/5] Resolving packages...
    success Already up-to-date.
    $ node ./node_modules/vscode/bin/install
    Detected VS Code engine version: ^1.15.0
    Found minimal version that qualifies engine range: 1.15.0
    Fetching vscode.d.ts from: https://raw.githubusercontent.com/Microsoft/vscode/8b95971d8cccd3afd86b35d4a0e098c189294ff2/src/vs/vscode.d.ts
    vscode.d.ts successfully installed!

and if I later try `yarn run vscode` I get

    yarn run v1.22.4
    warning pyls@0.0.1: The engine "vscode" appears to be invalid.
    $ npm run vscode:prepublish && VSCODE=$(which code-insiders || which code || echo echo ERROR: neither the code nor code-insiders vscode executable is installed); USER=dummy-dont-share-vscode-instance $VSCODE --user-data-dir=$PWD/.vscode-dev/user-data --extensionHomePath=$PWD/.vscode-dev/extensions --extensionDevelopmentPath=$PWD $* ..
    npm WARN lifecycle The node binary used for scripts is /tmp/yarn--1595363275707-0.8429611618286397/node but npm is using /usr/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.

    > pyls@0.0.1 vscode:prepublish /home/cesally/python-language-server/vscode-client
    > tsc -p ./

    /bin/sh: 1: /mnt/c/Users/cchak/AppData/Local/Programs/Microsoft: not found
    error Command failed with exit code 127.
    info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Any help resolving this would be appreciated.

ccordoba12:
No one is maintaining the VSCode extension at the moment, nor we have plans to fix that either, sorry (see my https://github.com/palantir/python-language-server/issues/194#issuecomment-646316798 about this).

However, you're welcome to submit a pull request to improve the situation and we'll be happy to merge it.

CHRC-Int:
Ok; thanks

PR #984, opened 2026-09-16
Document how to actually use the server with an editor

### LINKED ISSUE #194: Publish an official extension for vscode
Are there any plans to make the `vscode-client` available as an official extension, or is it designed for internal use only?

It could join the family of existing language server plugins like [Java](https://github.com/redhat-developer/vscode-java), [Scala](https://github.com/sbt/sbt/tree/1.x/vscode-sbt-scala) or [Rust](https://github.com/rust-lang-nursery/rls-vscode). There is already a Python plugin officially sponsored by Microsoft but it is significantly heavier in comparison.

Thoughts?

ZayanKhan-12:
## Description

Refs #194.

**This does not publish a VS Code extension, and is not a fix for #194 as titled.** @ccordoba12 already answered that in 2020 — "we don't have plans to create a VSCode extension for this server because most contributors use it for other editors and IDEs" — and that is not a contributor's call to overturn.

What the thread does contain is a request that was never answered. @munael, right after that reply:

> Can we add some documentation on enabling it for some IDEs? I assumed that installing the pip package would somehow register it for vscode to catch, but that doesn't seem to be the case.

and @Stercator in 2022:

> there's so little support on actually using it... without an actual integration, it stays just a dream

They are right that the docs do not cover this. The README explains how to `pip install` the package, and has a "Develop against VS Code" section for *developing the server*, but nothing anywhere says how a user connects it to an editor. #834, "Unable to run pyls with VS code", is still open.

## Changes

A "Using the server with an editor" section covering:

- The package installs a `pyls` executable; clients launch it and talk over stdin/stdout.
- `--tcp --host --port` exists for clients that cannot spawn the process.
- `pyls -v` is the first thing to reach for when an editor reports no results.
- Which LSP clients people commonly use per editor (Vim/Neovim, Emacs, Sublime Text, Kate).
- That server settings like `pyls.plugins.pydocstyle.enabled` are sent *by the client*, so they live in the client's config — which is not obvious and is the shape of several existing issues.

I verified both invocations rather than transcribing them from the argument parser: `pyls` over stdio and `pyls --tcp --port 8799` each complete an LSP `initialize` handshake and advertise the full capability set.

And a short **Visual Studio Code** subsection, placed where someone hunting for an extension will actually find it, recording that there is no official extension and no plans for one, linking your answer in #194. It also states plainly that `vscode-client/` is **not** that extension — it is unpublished, it is version `0.0.1` under an inherited publisher, and its own README says it launches a separate VS Code instance because it conflicts with other Python extensions. Someone arriving at that directory expecting a usable extension currently has no way to know this.

The second commit is separate and independent. The pydocstyle example in the Configuration section is fenced with markdown backticks inside an `.rst` file, so docutils reports `Possible title underline, too short for the title` and renders the snippet as ordinary text. `README.rst` is the `long_description` in `setup.py`, so that affects the PyPI page too. I noticed it because I link to that section. With it fixed, **README.rst parses with no warnings at all**. Drop that commit if you would rather keep this PR to one thing.

## Verification

- `README.rst` parses clean under docutils (no warnings, versus one before), and renders to HTML without error.
- `setup.py` still reads it as `long_description`.
- `pytest test/`: 12 failed, 99 passed, 8 skipped — **identical to the baseline on `develop`**, as expected for a docs-only change. Those 12 are pre-existing and track linter versions that moved on since 2020.
- `pycodestyle` clean, `pyflakes` unchanged (one pre-existing `_utils.py` warning).

## Notes

I have deliberately not named a specific third-party VS Code extension. Recommending one by name in your README implies a level of endorsement and review I am not in a position to offer, and those extensions come and go; describing the mechanism seemed more honest and more durable. Happy to add one if you have a preference.

If you would rather close #194 outright given the 2020 answer, this section at least means the next person to ask finds the answer in the README instead of eight years of issue history.

`CLAUDE.md` lives in my other open PR (#982) and is not duplicated here.


palantirtech:

